PR #4, opened 2015-11-20
Instrument Rework

OwenArnold:
Requirements/Design for the version 2.0 Geometry. At the time of writing this includes IDF and Geometry requirements. Maybe it would be wise to split this up.

Current requirements where largely non-functional, so I've split those out.

PR #5, opened 2015-11-30
Create AlgorithmUsageStatistics.md

NickDraper:
Design document for Algorithm usage statistics

### DIFF
--- a/Design/Usage/AlgorithmUsageStatistics.md
+++ b/Design/Usage/AlgorithmUsageStatistics.md
@@ -22,26 +22,35 @@ The contents of a Feature usage would be:
 * time - The datetime of execution start
 * duration - (optional) the execution duration
 * details - (optional) Further details - for algorithms this will be a properties string, or possibly the python string of the Algorithm call.
+* internal - true/false True if the interaction was not a direct response to user interaction (maps to alg.isChild()).
 
-These would be sent in addition to these fields in common with the startup Usage details:
+These would be sent in addition to several fields in common with the startup Usage details.  Take a look at the example json in the appendix for an example of a proposed message.
 
-* User ID hash
-* host ID hash
 
 ###New Code
 
-####API::UsageService
+####Kernel::UsageReporter
 
-We would create a new service class called Usage Service which would be responsible for collating, and sending
+We would create a new class called UsageReporter which would be responsible for collating, and sending
 all usage data.  This would centralise all the logic covering Usage Reporting including:
 
 1. Detecting if repoting is enabled
 1. Registering the startup of Mantid
 1. Sending Startup usage reports, immediately, and every 24 hours thereafter
 1. Registering feature usage, and storing in a feature usage buffer
-1. Sending Feature usage reports on application exit, and when the feature usage buffer is above a size threshold.
+1. Sending Feature usage reports on application exit, and when the feature usage buffer is above a size threshold.  This will need to be timed during development to ensure it does not add significantly to application shutdown.
 
 ###### Implementation Notes:
+This class will be owned and served out from the ConfigService throu a new UsageReporter method.  The ConfigService will also be responsible for the setup and lifetime of the UsageReporter class.  This is beacuse the UsageReporter needs to send a data package as part of it's destructor and needs the ConfigSerivice to still be available when it does that.
+
+The Usage report will have methods to :
+
+1. set and get an enabled status
+2. set the Application string, defaulting to "Mantidplot"
+3. flush the feature usage buffer
+4. set the interval for checking if time based jobs need to be done.
+
+Other Notes :
 
 1. Use Poco::Timer to handle the timed aspects of the class
   1. The 24 resend of startup data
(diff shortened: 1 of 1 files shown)


Design/Usage/AlgorithmUsageStatistics.md
```diff
@@ -0,0 +1,82 @@
...
+
+* type - Algorithm, Interface, Feature
+* name - Identifying name, for algorithms this would be Algorithm and version
+* time - The datetime of execution start
+* duration - (optional) the execution duration
+* details - (optional) Further details - for algorithms this will be a properties string, or possibly the python string of the Algorithm call.
+
+These would be sent in addition to these fields in common with the startup Usage details:
+
+* User ID hash
```
peterfpeterson: I think you would want to also have the mantid version and operating system as well

Design/Usage/AlgorithmUsageStatistics.md
```diff
@@ -0,0 +1,82 @@
...
+Clearly the django website https://github.com/mantidproject/webapp will need to be extended to accept feature usage reports.
+
+Initially we do not plan to define reports for the feature usage data, they will be queried direct from the database using SQL.
+Once a good understanding of the usefull aspects of the data clarifies we will look to integrating some reports into the API.
+
+###To Existing Code
+
+####API::FrameworkManager
+
+Currently the code to send usage reports is in the FrameworkManager (FrameworkManagerImpl::SendStartupUsageInfo).
```
peterfpeterson: The current place in `FrameworkManager` is the reason why the `application` is always `mantidplot`. The framework doesn't have a way to detect if it is being run from python or anything else.

Design/Usage/AlgorithmUsageStatistics.md
```diff
@@ -0,0 +1,82 @@
...
+####API::UsageService
+
+We would create a new service class called Usage Service which would be responsible for collating, and sending
+all usage data.  This would centralise all the logic covering Usage Reporting including:
+
+1. Detecting if repoting is enabled
+1. Registering the startup of Mantid
+1. Sending Startup usage reports, immediately, and every 24 hours thereafter
+1. Registering feature usage, and storing in a feature usage buffer
+1. Sending Feature usage reports on application exit, and when the feature usage buffer is above a size threshold.
```
peterfpeterson: There is already an issue with slow shut-down. Would this make it worse?

peterfpeterson:
This design needs more detail on what about the algorithms/interfaces is going to be sent over. I'm not currently persuaded that it couldn't be done using the existing server code as a `component`. The design would be modify only the client side.

Design/Usage/AlgorithmUsageStatistics.md
```diff
@@ -0,0 +1,82 @@
...
+
+###### Implementation Notes:
+
+1. Use Poco::Timer to handle the timed aspects of the class
```
peterfpeterson: Why not use/extend `MantidKernel::Timer`?

Design/Usage/AlgorithmUsageStatistics.md
NickDraper: I'll include some timings to take a look at that.

Design/Usage/AlgorithmUsageStatistics.md
NickDraper: Yes, actuually I'm including all of the information used in the Startup feature report in the message to the server (the share the same message header

Design/Usage/AlgorithmUsageStatistics.md
NickDraper: Yes,  I'm just porposing to tidy things up here, if we fix that later on it would be easy to add.

Design/Usage/AlgorithmUsageStatistics.md
peterfpeterson: You might be better off having the table of information of startup linked into the "algorithms" table.

Design/Usage/AlgorithmUsageStatistics.md
NickDraper: That just does wall clock timings, the Poco::Timer runs in a seperate thread and fires off callbacks on an interval.  Rather different, and it is already in use elsewhere.

Design/Usage/AlgorithmUsageStatistics.md
martyngigg: Could we have an example of the `json` payload that might be sent?

Design/Usage/AlgorithmUsageStatistics.md
martyngigg: With the new service we could simply have a field called `application` that can be set by a member function call and leave it to the given apps to set it. That way you don't have to infer anything. It'll be a bit like the Qt calls that you have to make to set the application name.

Design/Usage/AlgorithmUsageStatistics.md
```diff
@@ -0,0 +1,82 @@
...
+  1. The 24 resend of startup data
+  1. Checking every n minutes if the feature buffer above the threshold for sending.
+1. Registering Feature usage must be fast, just create the record and return.
+1. Do not lock the feature usage buffer for any longer than is absolutely necessary.
+1. Use a queue, rather than a vector for the feature usage buffer.
+1. Create multiple overloads for registerFeatureUsage() to make is easy for other developers, one should be specialised for algorithms.
+1. Internet calls should use the InternetHelper.
```
OwenArnold: I would prefer the usage statistics gathering to be a disconnected/passive as possible from the normal algorithm operation as not to affect stability. Possibly not in this issue, but in the near future, InternetHelper should be refactored to reduce the temporal coupling that can lead to instability.

Design/Usage/AlgorithmUsageStatistics.md
```diff
@@ -0,0 +1,82 @@
...
+Currently the code to send usage reports is in the FrameworkManager (FrameworkManagerImpl::SendStartupUsageInfo).  
+As described above this change would create a UsageService which would be responsible for collating, and sending
+all usage data.  As such I would propose to move the execution code from Framework Manager to the UsageService class, 
+and replace it with a single call to UsageService::Instance().registerStartup().
+This would also allow a single place to allow the additional logic to resend startup usage calls after 24 hours of constant running.
+Currently Long running instances cause the usage statistics of Mantid to underreport.
+
+###API::Algorithm
+
+We would add a call to UsageService::Instance().registerFeatureUsage(*this) within the Execute method of the Algorithm base class,
```
OwenArnold: Would it be possible to inject the usage service in the algorithm rather than to access the singleton in the method body? I would be in favour of being able to create and use an algorithm more separately from the rest of the framework if I wanted to.

Design/Usage/AlgorithmUsageStatistics.md
martyngigg: Agreed. To me this is all leading to having a separate service, as @OwenArnold has already suggested in the [DataProcessorAlgorithm UI design](https://github.com/mantidproject/documents/blob/master/Design/DataProcessorAlgorithmUI/DataProcessingUserInterface.md), that is responsible for running an algorithm and controls things such as:
 * handing workspaces over to the ADS;
 * enabling/disabling usage reporting;
 * + other stuff I haven't thought of

Design/Usage/AlgorithmUsageStatistics.md
NickDraper: The next version of this will have an adjustment, that the UsageService will actally be called UsageReporter and be owned and exposed by the Configservice.  This is only accessed by the algorithm base class, which has many many links to the rest of the framework.

OwenArnold: Indeed. Algorithm.cpp currently directly accesses singletons for the `AnalysisDataService` `AlgorithmManager` and `MemoryManager`. I don't yet have a picture of what the proposed architecture around this is, but it be nice to achieve more loose coupling that we currently have and I think that should be a consideration. If the new features could be designed in such a way that would be great. For example I may want to check that my algorithm consumes the usage reporting service in the right way without having to use the *real thing*.

martyngigg: I think some class diagrams of how things are related would be helpful.

peterfpeterson:
While I have a great desire to know what functionality is being used, I am concerned about the solution being heavy. Additional information about what is going to be captured would go a long way.

Design/Usage/AlgorithmUsageStatistics.md
```diff
@@ -0,0 +1,148 @@
...
+   "uid":"114ede53ec4e133a0d637f889ef8764d"
+} 
+```
+##Json for feature usage calls
+Note: This has an almost identical header than that of the startup calls.
+``` json
+{  
+   "ParaView":0,
+   "dateTime":"2015-11-30T16:22:35.270917000",
+   "features":[
```
peterfpeterson: I'm not comfortable with centrally logging all of the parameter values. It also gets "big" when the parameter is a `FitFunction` and useless when the parameter is a `PropertyManager`.

peterfpeterson:
I think the "Overview" (read: statement of problem) is spot on. Where I think this design needs more is in what the actual requirements are. The specific questions that aren't currently answered:
 1. Do we want automatic logging of algorithm/interface usage, or on demand?
 1. Do we want to link to the version of mantid, the userid, and host os? e.g. Do you want to know if `Rebin` is used more in v3.4.0 on mac or v3.5.1 on rhel? Do you want to know how many users use `Rebin`, or just how many times it was invoked?
 1. Do you want information about how the framework was started (e.g. `mantidplot` or `mantidpython` or ???)?
 1. Is it useful to know some/all of the parameters that we used in a algorithm? Same question for an interface? Same question for an ICAT query?
 2. Where is the collected information going to be presented? For example, would you add a line in the algorithm help like "84.2% of people use this algorithm" or "5.2% of mantid instances use this algorithm"? Or will there be a quarterly report presented to people? Or will be a report that is on the website (not in the help) and available to users?

martyngigg:
Agreed with @peterfpeterson, that information is missing here. @OwenArnold had a nice layout for this in https://github.com/mantidproject/documents/blob/DataProcessorAlgorithm/Design/DataProcessorAlgorithmUI/DataProcessingUserInterface.md#requirements

NickDraper:
I hope this answers your concerns and questions

Design/Usage/AlgorithmUsageStatistics.md
```diff
@@ -0,0 +1,185 @@
...
+
+This design document should be considered as an extension of the previous implemented design [MeasureUsageStatistics](MeasureUsageStatistics.md).
+
+#Requirements
+
+##Must
+
+1. Must record Mantid startup as is currently done.
+2. Must record algorithm usage, in a flexible way that allows ad-hoc queries to be performed.
+3. Must not significantly adversly affect Mantid interactive performance.
```
OwenArnold: For completeness. If we're going to say something about performance, I also think it would be worth specifying reliability (even though strictly they are both functional) as one of the must-have requirements. Basically the requirements should be that we do nothing to compromise normal operation of the algorithms or the framework. Sounds obvious but it should filter down into design considerations.

OwenArnold:
Apologies, if I missed this. But presumably we don't want to log algorithm usage from tests. How does the design handle this? This is probably covered in the existing functionality around start-ups, but I think it should be specified.

Other than that. Looks much improved.

NickDraper:
The unit tests and auto reduction is already handled using the UageReporting.Enabled user property which is always disabled in development, sys test and automated reduction.  That has been in place for the startup reporting for some time.

peterfpeterson:
@NickDraper This still has a major issue with security/privacy/size with logging all parameters of every algorithm. 

Separately, I think there should be an option to notify of startup, but not log all algorithms.

peterfpeterson:
Additional "must have": This cannot break current usage reporting system. This means that the underlying tables can change, but the `POST` for startup does not need to.

The example json violates "must have" #4 - no identifiable information. File paths will likely have user home areas which will identify the user. Similar argument for *anything* going into `SetSampleMaterial`.

Currently we are running on the free tier of openshift, this has implications to the database table design. There should be a representation/schema of what the database tables will look like. Information should be normalized. This is missing from the current design.

Logging child algorithm usage is more interesting. For example, `CalibrateRectangularDetector` calls  `FindPeaks` for each pixels separately (256x256x18 pixels on SNAP). `FindPeaks` calls `Fit` (multiple times) and `FindPeakBackground`. This would generate 10^7 "usages" in a single calibration. How will the design account for this?

For long-running execution, how/will we discern the difference between the startup and the 24h later?

Please leave `SendUsage` in place. While it will not be used automatically, there is a use case for adding it to autoreduction.

NickDraper:
@peterfpeterson  You raise good points about identifiable material and the volume of calibration calls.  Let's talk about that this morning/afternoon.  Please give me a call when you get in.

NickDraper:
Docunent adjusted to just store summary information only

peterfpeterson:
This was approved at the [2015-12-09 meeting](https://github.com/mantidproject/documents/blob/master/Project-Management/TechnicalSteeringCommittee/meetings/2015/TSC-meeting-2015-12-09.md).

PR #6, opened 2015-12-07
Create SANS beam centre finder design document

AntonPiccardoSelg:
We want to enhance the beam centre finder routine which is currently placed in the ISIS SANS GUI to allow for it to be used by other facilities and instruments.


OwenArnold:
Good start

Design/SANS/BeamCentreFinderDesign.md
```diff
@@ -0,0 +1,119 @@
+# Design Document for Beam Centre Finder
+
+### Current issue
+Currently the Beam Centre Finder(BCF) is heavily tied up with the ISIS instruments and the ReductionSingleton().
+ANSTO are looking currently for similar functionality but since their reduction workflow is not based on the
+ReductionSingleton() and they don't make use of the instrument class they cannot use the BCF directly. 
+This document lines out the required changes to make the BCF useable in a more general context which could 
+also be beneficial when moving away from the current Reduction-Step approach.
+
```
OwenArnold: Can you add a *Benefits* summary to justify the work. Include lower maintenace overhead for both facilities etc.

OwenArnold: Can you also add brief information about what approach the two facilities currently use. i.e why they are currently not compatible. For the ANSTO case, this is only planned so far.

Design/SANS/BeamCentreFinderDesign.md
```diff
@@ -0,0 +1,119 @@
...
+  to take something more general than a reducer and base our factory decision on this more general input. Once it
+  has decided for an ISIS instruement, we can make use of the current implementation.
+* The Beam Centre Logger should be injected and completly replaced by an Ansto implementation. We log into the ISIS
+  SANS GUI currently.
+
+### Things which need to be redesigned
+
+An abstract base class which can provide a more general interface could look like this
+```python
+class SANSBeamCentreFinderReductionWrapper(abc):
```
OwenArnold: You may call your *Adaptee* a wrapper. But as far as the client of the interfaces should be concerned, this isn't wrapping anything and they shouldn't need to know a lot about actual implementation details. Please drop the *Wrapper* term

Design/SANS/BeamCentreFinderDesign.md
```diff
@@ -0,0 +1,119 @@
...
+ANSTO are looking currently for similar functionality but since their reduction workflow is not based on the
+ReductionSingleton() and they don't make use of the instrument class they cannot use the BCF directly. 
+This document lines out the required changes to make the BCF useable in a more general context which could 
+also be beneficial when moving away from the current Reduction-Step approach.
+
+### Proposed solution
+In general terms we want to hide any solution-specific details in the BCF. Currently there is already a layer of 
+abstration which hides the instrument specific details, e.g. LARMOR uses an instrument rotation instead of a translation.
+A common generic interface should allow for the current implementation on the ISIS side as well as other 
+implementations to make use of the Beam Centre Finder logic.
```
OwenArnold: We should base the design on the current geometry. Since there is no way for an instrument component to know it's degrees of freedom. Individual implementations should have a **simple** way to specify how the banks could be moved. We should only deal with current cases of translation and rotation (one axis)

Design/SANS/BeamCentreFinderDesign.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Things which need to be redesigned
+
+An abstract base class which can provide a more general interface could look like this
+```python
+class SANSBeamCentreFinderReductionWrapper(abc):
+   ...
+
+   @abstractmethod
+   def initialize(xstart, ystart)
```
OwenArnold: Do we really need an * initialize* method?

Design/SANS/BeamCentreFinderDesign.md
```diff
@@ -0,0 +1,119 @@
...
+   Removes the EndNans from teh workspace
+   '''
+
+
+
+```
+
+
+### Things to clarify and potential difficulties
+*
```
OwenArnold: The interfaces may need to be assessable by algorithms under PythonInterfaces. So the interfaces probably shouldn't live under scripts.

picatess: I think that this should be possible, but the implementation of these interfaces should reside whereever the Reducer implementation lives, depending on a reduction-step- or reduction-workflow-based approach.

Design/SANS/BeamCentreFinderDesign.md
picatess: Calling it `getStartingValues` might be , but we do require intial values. In addition we should have a method `getStepSize`.

Design/SANS/BeamCentreFinderDesign.md
OwenArnold: Yes. The concrete implementations should be local.

OwenArnold: A class diagram for the proposed solution would be useful. Either using the dot format. See [here](https://github.com/mantidproject/documents/blob/master/Design/DataProcessorAlgorithmUI/DataProcessingUserInterface.md#current-structure) a hand drawing photographed would be just fine too.

Design/SANS/BeamCentreFinderDesign.md
```diff
@@ -0,0 +1,271 @@
+# Design Document for Beam Centre Finder
+
+### Current issue
+In order to calibrate the Beam Centre position with respect to the detectors, an accurate estimation estimation of the position needs to be performed. This functionality is provided by a Beam Centre Finder (*BCF*). 
+
+An implementation for this BCF already exists for ISIS SANS. This implementation is heavily tied up with the ISIS instruments and the specific way the data reduction is realized for ISIS SANS. ANSTO requires the same functionality as the current *BCF* provides. Unfortunately, it is not possible to directly use the *BCF*'s functionality with the ANSTO specific implementation of data reduction.
+
+The different approaches of ISIS and ANSTO are currently not compatible. They can be summarized as:
+
+* ISIS makes use of a so-called ReductionSingleton which has a series of registered reduction steps. During the reduction, the ReductionSingleton() carries state (mainly via the custom ISIS instrument classes). Intermediate results are stored in the Analysis Data Service (ADS).
```
OwenArnold: Might be worth specifying more about the ISIS Instrument. i.e. what it's responsibilities are.

Design/SANS/BeamCentreFinderDesign.md
```diff
@@ -0,0 +1,271 @@
...
+
+* ANSTO's reduction implementation is currently still work in progress, but will eventually make use of so-called Workflow-Algorithms which do not populate the ADS or carry state information. This will be a cleaner, more maintainable implementation which is easier to test and verify. It is desirable that in the long run ISIS SANS switches to such an implementation.
```
OwenArnold: DataProcessorAlgorithm would be more accurate than Workflow-Algorithms

Design/SANS/BeamCentreFinderDesign.md
```diff
@@ -0,0 +1,271 @@
...
+
+The static details of the current implementation (ie a class diagram) can be seen here:
+
+![alt text](beam_centre_finder_classes.png)
+
+
+The workflow of the current implementation is illustrated below. It is separated into two parts, one which is repsonsible for stepping through the parameter space
+![alt text](FindBeamCentre_py.png)
+
+and one which illustrates how the residuals -- a measure of our search quality -- are calculated. This is achieved by the *SeekCentre* operation:
```
OwenArnold: I don't think that any mention of `SeekCenter` is made before this. We do however have a `CenterFinder` in the diagram.

Design/SANS/BeamCentreFinderDesign.md
```diff
@@ -0,0 +1,271 @@
...
+
+![alt text](centre_finder_py.png)
+
+### Proposed solution
+In general terms we want to hide any solution-specific details in the *BCF*. Currently there is a layer of abstration which hides the instrument-specific details, e.g. LARMOR uses an instrument rotation instead of a translation. Extending this abstraction further to make the functionality also facility-agnostic will allow us to apply the *BCF* to any facility or instrument.
+
+One way of achieving this is a generic interface which is discussed below. In this case each instrument\facility needs to provide an implementation for the generic interface. This implementation is then injected into the BCF.
```
OwenArnold: Good

Design/SANS/BeamCentreFinderDesign.md
```diff
@@ -0,0 +1,271 @@
...
+and one which illustrates how the residuals -- a measure of our search quality -- are calculated. This is achieved by the *SeekCentre* operation:
+
+![alt text](centre_finder_py.png)
+
+### Proposed solution
+In general terms we want to hide any solution-specific details in the *BCF*. Currently there is a layer of abstration which hides the instrument-specific details, e.g. LARMOR uses an instrument rotation instead of a translation. Extending this abstraction further to make the functionality also facility-agnostic will allow us to apply the *BCF* to any facility or instrument.
+
+One way of achieving this is a generic interface which is discussed below. In this case each instrument\facility needs to provide an implementation for the generic interface. This implementation is then injected into the BCF. 
+
+Another, similar way, would be to use a Template Method approach. We would have to create a class-based version of the BCF. Each instrument/facility then implements the methods which need to be overwritten.
```
OwenArnold: Worth mentioning this Template Method option. But I might mention something about why Strategy injection would be superior.

OwenArnold:
I think this is pretty good. Once the fixes above have been made we should take it to the TSC.

OwenArnold:
@AntonPiccardoSelg I have merged this, but will still need to go via the TSC for approval.

PR #7, opened 2015-12-10
Live listener customisation

mantid-roman:
This is a PR for a design to provide more flexibility around ILiveData. ILiveData already uses the pluggable loading mechanism, so is a candidate for external facilities to use. However Mark Koennecke at the PSI has made some feature requests around this.

```
LiveListeners based on
instrument names falls short in two cases:
- when the instrument has two different detector systems with different interfaces which can be 
  selected by configuration. Case in point is BOA at SINQ which can have either a CCD-camera
  or a He3 EMBL detector attached. 
- Some ESS instruments will be designed multi method. IMHO, this is no good idea as such
  instruments are usually not optimal for any use case. But in this case you need to select if
  you want to listen to the image, or SANS, or powder data coming from the same instrument

I still think that a WatcherAlgorithm which can handle WS processing on WS updates from
another algorithm would be useful addition to Mantid. 
```


OwenArnold:
@peterfpeterson Your input is needed.

peterfpeterson:
The "dynamic properties" solution doesn't seem that far off. Unless I am mis-understanding the idea, it would be to have the type of `LiveListener` in the `Facilities.xml`, with the connection not made until `Exec` is called. A couple of wrinkles/good things:
 1. Most instruments have a fixed URL for live data (unique per instrument), it is preferred if the factory would return the correct `LiveListener` with the properties already correctly filled in for the instrument.
 2. Occasionally the URL changes or is different based on the subnet. This would allow for the user to change it at runtime.

OwenArnold:
@peterfpeterson Please review the PR again. I've added another option/solution.

peterfpeterson:
A question about "option 2": There must be a better name for the extra keyword than `hardware`. That doesn't seem to convey the correct concept. Do we want this added to the agenda at tomorrow's TSC?

OwenArnold:
@peterfpeterson. I didn't think that 'hardware' was such a bad name given that 'component' and 'type' have every meaning possible in Mantid. Having said that, I'd be more than happy to use something other than 'hardware' that might be suggested. When you're done with an initial review, I think this should go to the TSC.

martyngigg:
I'm not sure the requirements have been laid out clearly enough.

Is the issue that: an instrument may have multiple streams available and a user would like to select at run time, presumably in `StartLiveData`, what stream is used as the source?

If this is the case then extending the possibility for multiple listeners per instrument seems like the correct approach. Regarding the extra `hardware` keyword, what would this be used for? If the streams all have separate addresses then is this just used as an identifier for selection purposes? If it is for identification then would something simpler like `id` or `name` be better?

Should the design also make clear at this point that only a single connection per `StartLiveData` call can be made, i.e data will not be able to be accumulated from multiple streams here (unless that is an aim?)

peterfpeterson:
The interest at ORNL in this is that the DAS likes to change the IP of the live data stream (every time is the last time). This would be a mechanism for following those changes without releasing a new `Facilities.xml`.

martyngigg:
I see. So for that case the address needs to become a property that can be passed through `StartLiveData`?

peterfpeterson:
My impression was having dynamic properties and a factory would already cover it. A way to reuse what we have is to use the information in the `Facilities.xml` as defaults for the improved `StartLiveData` gui.

martyngigg:
Sounds okay to me.

OwenArnold:
@peterfpeterson @martyngigg. Thanks for input thus far. In this first round i've tried to address your concerns thus far. Please do raise any other issues.

OwenArnold:
@peterfpeterson: This is already assigned to you. I think you're the right man for the job. We need to choose the right option and put this through so that it can be given to a developer.

OwenArnold:
@peterfpeterson. As this has been waiting for several months, I'd like to get this out of stasis and give to Michael Hart to start work on it. Can you complete the review please.

peterfpeterson:
From information on slack today, it looks like @martyngigg started working on option 3. This does not seem like a bad way to go.

martyngigg:
My apologies. I hadn't realised I was stepping on the toes of this document as I'd forgotten about it.

My thinking now is that possibly some slight modifications to option 3 would be the best as ew have a new use case where the Live Listener needs to know the name of the instrument when it is connecting.

Here's my idea:

_We change the signature of the `ILiveListener::connect` method to take a single new `ConnectionArgs` type. One of the mandatory arguments would of course be the address and we can also add set the instrument name too. The class could inherit from `PropertyManager` to it can accept arbitrary named parameters in addition to some mandatory fields. The `FacilityInfo` parser can then be updated to pull out any additional `key=value` pairs and populate the `ConnectionArgs` `dict` to pass through `connect`._

_All that remains then I think is to expose further options on `StartLiveData` to be able pass over additional connection arguments at run time so that they can effectively fully overwrite the defaults._

Thoughts?

peterfpeterson:
That sounds about right. I assume that we'll keep the current url configuration in `Facilities.xml` and this is to add "extra" information at runtime.

Design/MoreFlexibleILiveListener.md
```diff
@@ -63,34 +64,89 @@ The general problem we are trying to solve is that the instrument is assumed to
...
+  </instrument>
+```
+
+#### Requirement 3
+
+Supporting direct LiveListener instantiation using a class name and connection string will require either modifying the existing `LiveListenerFactoryImpl::create` method, as per Option 2 above, or overloading it. The proposed solution is to overload it to provide an alternative while minimizing impact on existing code.
+
+The existing `create` method can be modified to function the same way it does now, but internally call the new `create` method, passing in the class name and address retrieved from Facilities.xml. `LiveDataAlgorithm` and `StartLiveData` (as well as its dialog) will need to be modified as well, to provide a user interface and call LiveListenerFactory accordingly.
+
+By adding a "Connection" group under the "Instrument" drop down box, to select an address / listener pair defined in the Facilities.xml using a drop down box, or to specify a custom address and listener, we could support both the general use case and allow for custom setups.
```
OwenArnold: I prefer the solution of having the two behaviours.

1. Filter listeners by facility + instrument
2. No filtering based on selected facility or instrument

Option 1. Is what we currently have with the addition of having the drop down selection (`StringListPropertyValidator`). Option 2 also requires the new property but in this case we do not use the current instrument or facilities xml to filter anything. We simply list what's in the `ILiveListener` `DynamicFactory`. We have a `Mantid.Properties` level switch for defining the behaviour and it defaults to filter (as in 1). 

I don't think that the Mantid user should have to know the live listener names and addresses (automatically listing them is preferred). Listing everything as part of behaviour 2, satisfies the requirement of authors of live listeners only having to provide the plugin and not also manage the facilities.xml file.

Design/MoreFlexibleILiveListener.md
```diff
@@ -63,34 +64,89 @@ The general problem we are trying to solve is that the instrument is assumed to
...
+
+
+Proposed Solution
+-----------------
+
+#### Requirement 1
+
+Dynamic properties for listeners were already supported previously, but this feature was removed in [PR #234](https://github.com/mantidproject/mantid/pull/234) in response to [Trac #11059](http://trac.mantidproject.org/mantid/ticket/11059) because it interfered with help generation and Python API support.
+
+Reverting this PR will satisfy this requirement, but we will need to deal with the help and Python issues in a different way. This is beyond the scope of this design and can be handled separately.
```
OwenArnold: @martyngigg. While i think this change set should be reverted. I think that @MikeHart85  has a good point about the way that the python functional syntax for the `StartLiveData` is created. Do you have any suggestions to get around this?

Design/MoreFlexibleILiveListener.md
MikeHart85: That's how I imagined it at first as well, but I don't think that will work because you don't really filter listeners but rather connections. Multiple connections for the same instrument can use the same listener class (but different addresses). At least that's what it looks like based on the xml in Partial Solution 1.

To support this in the user interface, we would have to add a drop down box that lists, for example, "instrument", "detector1" and "bank1" when instrument "LET" is selected.

For Option 2, the user will wants to specify both a listener class and an address string. Even if the address is moved to "Listener Properties", they still need to specify a listener class name rather than a connection name. This doesn't fit with Option 1.

So my thought was to also list "[Custom]" along with "instrument", "detector1", "bank1" and have an interface that looks something like:

```
 Instrument       [ Dropdown Box    |V]

-------- Connection Properties ---------
| Connection      [ Dropdown Box   |V] |
| Listener Class  [ Dropdown Box   |V] |
| Address String  [ Text Box         ] |
----------------------------------------
```
Where:
* When you select an Instrument, Connection is populated with connection names and preset based on the `default_name` attribute of the `livedata` tag
* When you select a Connection (or as a result of the above), the Listener Class and Address String are filled in for you based on that connection's attributes in facilities.xml
* If you modify Listener Class or Address String, Connection is set to "[Custom]"

What are your thoughts on this?

OwenArnold: Yes you're right with some of this. It always catches me out that the items in the `DynamicFactory` are their own factory. We really need a solution that doesn't require modifying the `facilities.xml` file. I haven't considered your proposal thoroughly yet though.

martyngigg:
Thanks for tidying this up @MikeHart85. It's now much clearer what the issues are and how the proposed solution will work. I'm happy.

Design/MoreFlexibleILiveListener.md
OwenArnold: @MikeHart85 you update the document with martyn's suggestion. This is part of the scope of the design. But we may choose to split the implementation into several issues/prs.

Design/MoreFlexibleILiveListener.md
```diff
@@ -0,0 +1,180 @@
...
+
+Order of Connection
+-------------------
+
+As used at the moment, `LiveDataAlgorithm` has a gateway function called [getLiveListener](https://github.com/mantidproject/mantid/blob/master/Framework/LiveData/src/LiveDataAlgorithm.cpp#L186:L197), which forces `LiveListenerFactory::create()` to return a `ILiveDataListener` on which connect has been called because the connect argument is set to true. `getLiveListener` also starts the listener. If you don't have enough information to start the connection at that point, the connection will simply fail.
+
+
+Questions
+---------
+Mark mentioned that we are missing a Watcher Algorithm. It's unclear what is required that's not already handled by `MontiorLiveData` and `LoadLiveData` via `StartLiveData`?
```
peterfpeterson: I believe that this is exactly what `MonitorLiveData` was intended for.

peterfpeterson:
It isn't clear which option is being implemented from the design document which should be able to stand on its own without the conversation above. Clarify that point then I'm happy to add my initials to endorse the document.

Design/MoreFlexibleILiveListener.md
MikeHart85: I've removed the statement that this is out of scope, and replaced it with Martyn's suggestion (adjusted for StartLiveData arguments).

MikeHart85:
@peterfpeterson 
I've added a small paragraph at the top of Proposed Solution to explain how the section relates to the options above. Is this what you had in mind?

PR #8, opened 2016-02-02
Simplifying calling child algorithms

peterfpeterson:


OwenArnold:
It would be nice if the `exec..` methods all have all had the ability to explicitly return whatever they've created. I think this is where you are going with the design.

It looks like you're now proposing switching the default so that Algorithms are run with child=true as the default?

Design/Algorithm_ease_of_use.md
```diff
@@ -0,0 +1,88 @@
...
+like
+
+```C
+double tolerance = getProperty("CompressTolerance");
+m_outputEW = boost::dynamic_pointer_cast<EventWorkspace>(m_outputW);
+if ((m_outputEW) && (tolerance > 0.)) {
+  g_log.information() << "running CompressEvents(Tolerance=" << tolerance
+                      << ")\n";
+  API::IAlgorithm_sptr compressAlg = createChildAlgorithm("CompressEvents");
+  compressAlg->setProperty("InputWorkspace", m_outputEW);
```
OwenArnold: I have a problem with member variables `m_` i.e. `m_outputEW` being used from any dynamic property. exec should be stateless in the context of the Algorithm object unless it's absolutely necessary to cache things for performance reasons. Can we consider fixing this somehow, or at least making it harder to do. How about `exec` becoming a free function called by the algorithm, where we pass in all the Properties needed?

peterfpeterson:
I'm not changing the default. You'll notice that the new methods are both called `execAsChild`. To be consistent they should probably be called `executeAsChild`. There doesn't appear to be a generic way (for the second alternative) to return the output property(s).

AndreiSavici:
I think alternative 2 is better. I don't see how in alternative 1 you can cope with multiple optional parameters.

Design/Algorithm_ease_of_use.md
peterfpeterson: This was cut-and-pasted from [AlignAndFocusPowder](https://github.com/mantidproject/mantid/blob/266f2cd000593dd78f061bf6c3c2227857ff169f/Framework/WorkflowAlgorithms/src/AlignAndFocusPowder.cpp#L364). I wanted to make an example from something found in the wild.

peterfpeterson:
@AndreiSavici Is the added text a good enough explanation?

Design/Algorithm_ease_of_use.md
OwenArnold: Sure, and I know that you deliberately chose a less than perfect implementation to illustrate the point. As I was looking at it I was wondering if it actually makes sense for us to think of Algorithms as classes. They're more like functions, and I was wondering if taking a functional approach to `exec` and below would be beneficial.

peterfpeterson: That is *really* what I was trying to figure out how to propose. If you have a way to do it, add it in as "Alternative 3".

OwenArnold: I think you're nearly there with "Alternative 2". You clearly need to call the non-members from the algorithm methods `execAsChild` or whatever. At that point, properties should be expanded out so that the typed values of each property can be passed as arguments. I'm not sure the best way to handle return types.

peterfpeterson:
Turns out I mis-understood Andrei's comment. Alternative 1 is (unfortunately) a hand written solution (per algorithm) that doesn't generically apply.

MichaelWedel:
I like alternative 2 in that it behaves a bit like `kwargs` in Python, which is very useful. It's the callers responsibility to provide the correct input properties. How would the map be declared with mixed property types?

peterfpeterson:
I should have written a commit message. I just added a (bad) implementation of `executeAsChildAlg` for alternative 2.

quantumsteve:
I think alternative 2 would look a lot like [PR #15104](https://github.com/mantidproject/mantid/pull/15104), where I updated how we construct many STL containers.

martyngigg:
I'm keen on alternative 2 as well for the same reason as @MichaelWedel.

I agree with @OwenArnold that we need some way to access output values. As I see it the only thing we can do is either return the algorithm pointer or return some container of the output properties.

MichaelWedel:
Is it possible to return some kind of map container where the caller specifies what properties to extract from the algorithm via a string list?

```
auto result = execAsChild("algorithm", {"InputWorkspace", someWorkspacePointer}, {"OutputWorkspace"});
```

I guess the problem is that you don't know the types beforehand, something like `QVariantMap` would be useful in this case.

Design/Algorithm_ease_of_use.md
```diff
@@ -0,0 +1,100 @@
...
+```C
+double tolerance = getProperty("CompressTolerance");
+m_outputEW = boost::dynamic_pointer_cast<EventWorkspace>(m_outputW);
+if ((m_outputEW) && (tolerance > 0.)) {
+  g_log.information() << "running CompressEvents(Tolerance=" << tolerance
+                      << ")\n";
+  API::IAlgorithm_sptr compressAlg = createChildAlgorithm("CompressEvents");
+  compressAlg->executeAsChildAlg({"InputWorkspace", m_outputEW},
+                           {"OutputWorkspace", m_outputEW},
+                           {"Tolerance", tolerance});
```
FedeMPouzols: I like the direction in which Alternative 2/3? is going.
Now what will code complexity tools say about this? They will disklike the type of calls that we will have here, with many properties passed in a single method call.
Maybe algorithms should have an input dictionary and an output dictionary of properties?

AndreiSavici:
@peterfpeterson ?

PR #9, opened 2016-02-03
Design for powder diffraction calibration

peterfpeterson:


Design/PDcalibration.md
```diff
@@ -0,0 +1,145 @@
...
+     ([docs](http://docs.mantidproject.org/v3.5.1/algorithms/GetDetOffsetsMultiPeaks-v1.html)
+     and
+     [github](https://github.com/mantidproject/mantid/blob/v3.5.1/Code/Mantid/Framework/Algorithms/src/GetDetOffsetsMultiPeaks.cpp))
+     is representative.
+
+   * Algorithm that performs bank-by-bank calibration to generate a
+     multiplicative factor for each summed
+     spectra. [EnggCalibrate](https://github.com/mantidproject/mantid/blob/v3.5.1/Code/Mantid/Framework/PythonInterface/plugins/algorithms/EnggCalibrate.py)
+     is representative of this approach.
+
```
FedeMPouzols: How often do they do the long run for pixel-by-pixel calibration on NOMAD?
With ENGIN-X the very long calibration run is done every ~[3, 5] years only. Then the bank-by-bank calibration runs are typically done once or twice per cycle or few months.
For ISIS powder diffraction instruments the current practice is apparently to re-use .cal files for a few cycles or around one year. I'd suspect this must be similar as for NOMAD. But we don't have this properly under control at present.

Design/PDcalibration.md
```diff
@@ -0,0 +1,145 @@
...
+   equation above. Since it is a scalar, it can be calculated as the
+   average `need_name` for each of the peaks in the observed
+   specturm. This is the 1-parameter version of the Gauss-Markov
+   theorem.
+
+Updating the `idf`
+------------------
+
+Most of this has been done. It involves merging
+[AlignComponents](https://github.com/mantidproject/mantid/compare/master...rosswhitfield:AlignComponent)
```
FedeMPouzols: Aha, if I'm following correctly, this new algorithm is a much more general and fancier way of doing what for example for EnginX is done with a simple pixel-by-pixel formula (as they only update the L2 of the individual detectors using the DIFC's). Then you just need the trivial `ApplyCalibration`. 

When this is implemented it might be a good idea to add some notes in these algorithms to clarify
how `ApplyCalibration`, `AlignComponents` and friends interplay, differ and/or overlap.

Design/PDcalibration.md
```diff
@@ -0,0 +1,145 @@
...
+by simple multiplication, e.g. `DIFC_new = DIFC * need_name`. (**OTHER
+EQUATION SHOULD CHANGE TO MAKE THIS CORRECT**)
+
+1. The first three steps are the same as the pixel-by-pixel except for
+   calculations and parameters are in d-spacing.
+
+1. Calculate the calibration constants, `need_name`, which satisfy the
+   equation above. Since it is a scalar, it can be calculated as the
+   average `need_name` for each of the peaks in the observed
+   specturm. This is the 1-parameter version of the Gauss-Markov
```
FedeMPouzols: Typo: `specturm`

peterfpeterson:
I'll answer most of the inline comments at once:
* POWGEN recalibrates every run cycle. This is especially bad since they run a new calibration in each sample environment as they all put the sample in slightly different (although consistent for the environment) places.
* NOMAD recalibrates for every experiment (read: every 1-3 days).  Even with this, they are having some issues getting the sample to go to the same place as the calibration run. The consistency is being addressed by hardware changes.
* VULCAN has their own calibration techniques that mantid can read in. I think it happens similar to ENGINX.
* `AlignComponents` is approaching a state where it can be merged into master. There is some missing functionality (it only tries to fit `DIFC` and just ignores other parameters), but it is already quite usable. You can merge it into your own code to see what it does for other instruments.

Design/PDcalibration.md
rosswhitfield: `AlignComponets` doesn't apply a corrections per detector but tries to correct the position/rotation of a detector bank as a whole component.

FedeMPouzols: Yes. For ENGINX our scientist are happy with just the per-bank difc, tzero parameters that they can pass on to GSAS. I think they had never considered the possibility of having an instrument geometry corrected with the per-bank calibration in Mantid. I'll check with them. They might find this new algorithm useful.

FedeMPouzols:
One question that came up while chatting about this over here: Will AlignComponents work well with all types of detectors (non-rectangular)?

peterfpeterson:
We've used it with `RectangularDetector` and 8-packs of LPSDs. It requires that the thing you align is an instrument component.

PR #10, opened 2016-02-05
Algorithm refactoring

SimonHeybrock:
Discussion of algorithm refactoring.

### DIFF
--- a/Design/algorithm_refactoring.md
+++ b/Design/algorithm_refactoring.md
@@ -84,6 +84,34 @@ Things to consider in the design:
 
 -  We need to provide means to deal with several in- and outputs, e.g., in the common case of an input and output `MatrixWorkspace`, where we need x, y, and error data for both input and output. There are also cases that are more complex than that
 
+- Why pass both a workspace and a getter to transform instead of just a function binding to the correct workspace and getter?
+  - The `PARALLEL` macros need the workspace to check whether or not parallel execution is possible.
+  - We may want to call things like `workspace.clearMRU();`.
+  - The users of `transform` would need to use `std::bind`, which is a bit nasty when binding to overloaded functions, e.g., the `const` overloads we have for getters on our workspaces: Basically you need to cast explicitly, e.g.,
+
+    ```cpp
+    std::bind((double &(std::vector<double>::*)(size_t))&std::vector<double>::operator[], &data, _1)
+    ```
+
+    Note that it is not possible to use implicit type conversion in combination with templates. Suppose we modify `transform` to take a `std::function<T&(size_t)>` as a getter. We then **cannot** use
+
+    ```cpp
+    transform(/*...*/,
+              std::bind((double &(std::vector<double>::*)(size_t))&std::vector<double>::operator[],
+                        &data,
+                       _1), /*...*/);
+     ```
+
+     but need to take a detour via a variable,
+
+    ```cpp
+    auto func = std::bind((double &(std::vector<double>::*)(size_t)) &
+                                      std::vector<double>::operator[],
+                                  &data, _1);
+    transform(/*...*/, func, &data, _1), /*...*/);
+       ```
+
+
 
 Furthermore, we would probably want to provide overloads for supporting the following cases:
 
@@ -158,7 +186,8 @@ outputWS = inputWS->clone();
 Apart from the bits covered by the discussion above on casting and creating output event workspaces, there are also others:
 
 - Dealing with the case `inputWS == outputWS`, which is particularly bad in the case of `EventWorkspace`, since it requires even more casting.
-- Setting the output workspace property,
+- Setting the output workspace property.
+
   ```cpp
   this->setProperty("OutputWorkspace", outputWS);
    ```
--- a/Design/algorithm_refactoring.md
+++ b/Design/algorithm_refactoring.md
@@ -157,7 +157,7 @@ outputWS = inputWS->clone();
 
 Apart from the bits covered by the discussion above on casting and creating output event workspaces, there are also others:
 
-- Dealing with the case `inputWS == outputWS`, which is particularely bad in the case of `EventWorkspace`, since it requires even more casting.
+- Dealing with the case `inputWS == outputWS`, which is particularly bad in the case of `EventWorkspace`, since it requires even more casting.
 - Setting the output workspace property,
   ```cpp
   this->setProperty("OutputWorkspace", outputWS);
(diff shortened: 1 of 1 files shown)


Design/algorithm_refactoring.md
```diff
@@ -0,0 +1,209 @@
...
+If it were not a member we would need some way of passing all these things into our `transform`, which is cumbersome.
+
+In the simplest case (in-place modification of a single field), this could be implemented along the lines of this:
+
+```cpp
+// Note that we need the explicit syntax ELEM (WS::*getter)(IDX), else the compiler cannot resolve (const) overloads
+template <class WS, class OP, class ELEM, class IDX>
+void transform(WS &workspace, ELEM (WS::*getter)(IDX), const OP &operation) {
+  int length = workspace.getNumberHistograms();
+  auto progress = Kernel::make_unique<API::Progress>(this, 0.0, 1.0, length);
```
AndreiSavici: Remember that for more complicated algorithms the loop over spectra might not take all the time. So one needs to pass minimum and maximum for progress (we can make those to default to 0 and 1)

AndreiSavici:
We should also make sure that any refactoring is compatible with the proposed new workspace structure.  @martyngigg is going to post a summary of the discussions from the developer workshop.

martyngigg:
In general I like the proposed `transform` solution. It would have been nice to have it as a non-member helper but I can see that is not going to be feasible with the current `PARALLEL` macros and progress implementation. Is the solution for other combinations of input/output going to be using overloads?

As an aside, while looking at the current `PARALLEL` macros I'm not sure why `m_parallelException` is a member variable - it doesn't look like it needs to be. They are also declared down in "Kernel/MultiThreaded.h" but are tied to how the algorithms work so as an aside I would like to move them next to the `Algorithm` header or defined in it. I'm also not keen on how the current `cancel` checks work. If you're in an openmp loop the interruption point does nothing and the cancel checks are performed in the macro - it's terribly inconsistent. Given that the `PARALLEL_START` & `PARALLEL_END` macros check for thrown exceptions I don't see why `interruption_point` can't be the same for both cases (espcially after https://github.com/mantidproject/mantid/pull/15148).

SimonHeybrock:
@martyngigg Yes, the plan is to have overloads for, e.g., one input workspace and one output workspace. When a transform needs multiple values from a workspace or has multiple outputs, a `std::tuple` of getters can be passed in.
This actually turns out to be a bit tricky: My initial plan was to have sth. like `transform(ws, std::make_tuple(&MatrixWorkspace::readX, &MatrixWorkspace::readY), ...)`. But we run into the same problem as mentioned with `std::bind`. `make_tuple` uses a forwarding reference (aka universal reference), and they don't work with overloads (the compiler cannot tell which one we want). So we would either have to explicitly specify the template arguments to `make_tuple` (and maybe cast), or use `std::function`. The latter is much nicer for the users, we could just provide the commonly used functions somewhere.

peterfpeterson:
Another thing that this possibly opens up for us is to have the functionality exposed to python so those algorithms could be "simpler" (use mostly defines how to transform a spectrum) and automatically get multi-threading.

Design/algorithm_refactoring.md
```diff
@@ -0,0 +1,209 @@
...
+- `Algorithm::Execute()` calls `getInputWorkspaceForVisitorPattern()` to obtain a workspace `ws` and then calls `ws.apply(*this);`
+- All algorithms implement `apply` for various workspace types.
+
+The last bullet is where it all starts to fall apart, there is a fatal flaw in this mechanism: Inheritance does not go well with overloading. For the visitor pattern to work we would need to implement `Algorithm::apply` for **all workspace types**. What we would actually like to do is, e.g., implement `Rebin::apply(const MatrixWorkspace &ws)` and `Rebin::apply(const EventWorkspace &ws)`. Calling the algorithm with a `Workspace2D` will then **not** resolve to `Rebin::apply(const MatrixWorkspace &ws)` but `Algorithm::apply(const Workspace2D &ws)`.
+
+The conclusion is that currently there is **no reasonable way for using the visitor pattern** in this case.
+
+- Are there other ways to solve the some problem?
+- Reconsider the visitor pattern when redesigning workspaces.
+
```
OwenArnold: Good summary of visitor. The overloading is the issue! Algorithm would need to be aware of all workspace types. I think this problem may have been tacked in the "Modern C++ Design" book. I think there's a chapter towards the end on this.

Design/algorithm_refactoring.md
```diff
@@ -0,0 +1,209 @@
...
+- The `PARALLEL` macros access `m_cancel`, `m_parallelException` (which, however, could be made local), and `g_log` (which may not be necessary, strictly speaking).
+- We would like to move the `Progress` code into our `transform`, so we need the `this` pointer.
+
+If it were not a member we would need some way of passing all these things into our `transform`, which is cumbersome.
+
+In the simplest case (in-place modification of a single field), this could be implemented along the lines of this:
+
+```cpp
+// Note that we need the explicit syntax ELEM (WS::*getter)(IDX), else the compiler cannot resolve (const) overloads
+template <class WS, class OP, class ELEM, class IDX>
```
OwenArnold: I quite like the policy based class design approach here. I think this sort of approach would be necessary for runtime performance.

Design/algorithm_refactoring.md
FedeMPouzols: If only we had the new re-designed workspaces: http://www.mantidproject.org/DevWorkshop2016_Thu_small_group.

Design/algorithm_refactoring.md
SimonHeybrock: I also fear that the `transform` templates will increase build times, yes.

How about putting the `transform` in a child class of `Algorithm`? All classes that do loop over spectra can then inherit from that.

Design/algorithm_refactoring.md
SimonHeybrock: @OwenArnold If also found a chapter in "More effective C++" (Item 31: Making functions virtual with respect to more than one object), but it is quite ugly, and actually does not seem solve the problem.

Design/algorithm_refactoring.md
martyngigg: I think this is a good plan. Some algorithms, like `DownloadInstrument` certainly doesn't want any of this stuff. It would be good to try and make the `Algorithm` class simpler and move things out.

peterfpeterson:
@quantumsteve pointed out to me yesterday that this example more closely follows `std::for_each` than `std::transform`. I don't know if we want both, but the name should match the functionality.

AndreiSavici:
@peterfpeterson ?

PR #11, opened 2016-02-08
Codebase restructure

martyngigg:
This document is meant to start the discussion of how to improve the structure of the code base based on the first discussion session at the 2016 developer meeting.

### DIFF
--- a/Design/CodeReorg/CodeReorg.md
+++ b/Design/CodeReorg/CodeReorg.md
@@ -20,8 +20,6 @@ The current structure of the codebase has remained largely unaltered since the p
 
 ## Proposed Solution ##
 
-## Solution Details ##
-
 Simple tasks:
 
 - Remove `UserAlgorithms` from being built and shipped with Mantid. This will include removing the development headers from the Windows package. Keep the source code but rename to `Examples`
@@ -32,32 +30,65 @@ Simple tasks:
 
 ### Root-directory Structure ###
 
-The proposed root-directory structure is as follows:
+The proposed directory structure (without the framework detail) is as follows:
 
-    -/
-     -cmake/
-     -devtools/
-       -buildconfig
-       -tools
-     -docs/
-     -framework/
-     -instrument/
-     -qt/
-       -applications/
-         -mantidplot/
-	   -qtpropertybrowser
-	   -paraviewext
-     -resources/
-        -fonts
-        -images
-        -properties
-     -scripts/
-     -testing/
-       -data
-     -thirdparty/
+	mantid.git
+	|-- cmake
+	|-- devtools
+	|   |-- buildconfig
+	|   |-- tools
+	|-- docs
+	|   |-- doxygen
+	|   |-- sphinx
+	|-- framework
+	|-- instrument
+	|-- qt
+	|   |-- applications
+	|   |   |-- mantidplot
+	|   |   |-- vsi
+	|   |-- qtpropertybrowser
+	|   |-- paraviewext
+	|   |   |-- api
+	|   |   |-- paraviewplugins
+	|   |-- widgets
+	|   |   |-- resources
+	|-- resources
+	|   |-- fonts
+	|   |-- images
+	|   |-- properties
+	|-- scripts
+	|-- testing
+	|   |-- data
+	|-- thirdparty
 
 ### Framework ###
 
-Notes:
+The proposed directory structure for the framework package is as follows:
 
-Old geometry should go in a `legacy` directory
+	framework
+	|-- catalog
+	|-- common
+	|   |-- algorithms
+	|   |-- constants
+	|   |-- core
+	|   |-- dataobjects
+	|   |-- units
+	|-- geometry
+	|   |-- legacy
+	|-- io
+	|-- mpi
+	|-- optimization
+	|-- neutron
+	|   |-- diffraction
+	|   |-- inelastic
+	|   |-- crystallography
+	|   |...
+	|-- muon
+	|-- remote
+	|   |-- live
+	|   |   |-- common
+	|   |   |-- ess
+	|   |   |-- isis
+	|   |   |-- sns
+	|   |-- cluster
+	|   |-- scriptrepository
--- a/Design/CodeReorg/CodeReorg.md
+++ b/Design/CodeReorg/CodeReorg.md
@@ -111,3 +111,8 @@ Notes:
 
 - each subdirectory will have its own `testing` directory
 - the library names will be formed from concatenating directory names togther, e.g. `common/core` would produce `libmantid-common-algorithm.so`.
+
+Sandbox
+-------
+
+To aid in discussing the structure I have created a test repository to play around with the layout. Please see [here](https://github.com/martyngigg/codereorg).
--- a/Design/CodeReorg/CodeReorg.md
+++ b/Design/CodeReorg/CodeReorg.md
@@ -73,17 +73,17 @@ The proposed directory structure (without the framework detail) is as follows:
 The proposed directory structure for the framework package is as follows:
 
 	framework
-	|-- catalog
+	|-- catalog: ICAT layer
 	|-- common
-	|   |-- algorithm
-	|   |-- constants
-	|   |-- core
-	|   |-- datamodel
+	|   |-- algorithm: algorithm interface + impl classes (dataprocessor), property manager, basic properties
+	|   |-- constants: maths: e.g. 2*pi, physical: hbar, neutronic: neutron mass
+	|   |-- core: System.h, MantidVersion.h ...
+	|   |-- datamodel: workspaces, axes, validators, workspace property
 	|   |-- units
 	|-- geometry
 	|   |-- legacy
-	|-- io
-	|   |--
+	|-- io: file loader interface, file searching, fileproperty
+	|   |-- nexus: contents of our nexus directory
 	|-- mpi
 	|-- optimization
 	|   |-- common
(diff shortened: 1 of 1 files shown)


Design/CodeReorg/CodeReorg.md
```diff
@@ -0,0 +1,113 @@
...
+	|   |-- tools
+	|-- docs
+	|   |-- doxygen
+	|   |-- sphinx
+	|-- framework
+	|-- instrument
+	|-- qt
+	|   |-- applications
+	|   |   |-- mantidplot
+	|   |   |-- mdviewer (formerly standalone VSI)
```
OwenArnold: I like the `mdviewer` rename

Design/CodeReorg/CodeReorg.md
```diff
@@ -0,0 +1,113 @@
...
+	|   |   |-- instrumentview
+	|   |-- resources
+	|-- resources
+	|   |-- fonts
+	|   |-- images
+	|-- scripts
+	|-- testing
+	|   |-- data
+	|-- thirdparty
+
```
OwenArnold: Not clear where custom interfaces would go

Design/CodeReorg/CodeReorg.md
```diff
@@ -0,0 +1,113 @@
...
+	|   |   |-- sns
+	|   |-- cluster
+	|   |-- scriptrepository
+	|-- resources
+	|   |-- properties
+
+Notes:
+
+- each subdirectory will have its own `testing` directory
+- the library names will be formed from concatenating directory names togther, e.g. `common/core` would produce `libmantid-common-algorithm.so`.
```
OwenArnold: Not clear where algorithms would sit in this, and more importantly the base classes for the `runnable` things

OwenArnold:
As a side-note, I think it would be helpful to somehow disambiguate the concept of `core` which has thus far been used (frequently and loosely) in discussions to describe:

* Things that you should touch and are OK to do so because there's sufficient testing
* Things that you should never touch because they're too important 
* Algorithms that are upstream of other algorithms 
* Parts of the system you always want to ship
* Code that somehow has to undergo greater QA that the rest

Design/CodeReorg/CodeReorg.md
```diff
@@ -0,0 +1,113 @@
...
+	|   |-- algorithm: algorithm interface + impl classes (dataprocessor), property manager, basic properties
+	|   |-- constants: maths: e.g. 2*pi, physical: hbar, neutronic: neutron mass
+	|   |-- core: System.h, MantidVersion.h ...
+	|   |-- datamodel: workspaces, axes, validators, workspace property
+	|   |-- units
+	|-- geometry
+	|   |-- legacy
+	|-- io: file loader interface, file searching, fileproperty
+	|   |-- nexus: contents of our nexus directory
+	|-- mpi
```
SimonHeybrock: I think in the future we will have MPI (related) code in various algorithms, without the current clear separation between `Algorithm` and `MPIAlgorithm`. It may therefore not be clear what would go into this directory. But I guess we can just keep it for now, and see later if it will still be necessary.

Design/CodeReorg/CodeReorg.md
martyngigg: @OwenArnold What do you mean by `runnable` things? My first thought was things like `StartLiveData` would be in `live/common`.

Design/CodeReorg/CodeReorg.md
martyngigg: That's a good point. Maybe we just don't need the distinction at all. I guess it's more of an implementation detail really...

Design/CodeReorg/CodeReorg.md
```diff
@@ -0,0 +1,113 @@
...
+	|   |   |-- common
+	|   |   |-- paraviewplugins
+	|   |-- widgets
+	|	|   |-- qtpropertybrowser
+	|   |   |-- instrumentview
+	|   |-- resources
+	|-- resources
+	|   |-- fonts
+	|   |-- images
+	|-- scripts
```
FedeMPouzols: This `scripts` would be the same as our current scripts? This is probably a can of worms, but I think that this currently contains too heterogeneous things. There are some files that are sort of custom (Qt) interfaces, reduction scripts; some of them are tested, some are not, etc.

Design/CodeReorg/CodeReorg.md
```diff
@@ -0,0 +1,113 @@
...
+	|-- testing
+	|   |-- data
```
FedeMPouzols: Would this `testing` also include performance, system, and tools, as in our current `Testing`?

Design/CodeReorg/CodeReorg.md
FedeMPouzols: :+1:

Design/CodeReorg/CodeReorg.md
martyngigg: That's a good point. I would at least like to factor out the interface stuff and put that along side the C++ interfaces. I don't see why we should differentiate

Design/CodeReorg/CodeReorg.md
martyngigg: That was my intention at the moment

Design/CodeReorg/CodeReorg.md
```diff
@@ -0,0 +1,113 @@
...
+	|-- optimization
+	|   |-- common
+	|   |-- costfunctions
+	|   |-- minimizers
+	|   |-- models
+	|-- neutron
+	|   |-- diffraction
+	|   |-- inelastic
+	|   |-- crystallography
+	|   |...
```
FedeMPouzols: Where would the Python algorithms and functions go? We currently have them in `Framework/PythonInterface/plugins/algorithms` and `Framework/PythonInterface/plugins/functions`. This includes many workflow algorithms which would find a proper place to live in their technique subdirectories. Probably they should go with their C++ siblings, unless that creates too many issues with cmake and scripts.

Design/CodeReorg/CodeReorg.md
FedeMPouzols: Sounds good, only that it is not very explicit in the diagram.

Design/CodeReorg/CodeReorg.md
```diff
@@ -0,0 +1,113 @@
...
+	|-- io: file loader interface, file searching, fileproperty
+	|   |-- nexus: contents of our nexus directory
+	|-- mpi
+	|-- optimization
+	|   |-- common
+	|   |-- costfunctions
+	|   |-- minimizers
+	|   |-- models
+	|-- neutron
+	|   |-- diffraction
```
FedeMPouzols: Does diffraction need one more level: common, SXD, powder, engineering, powder, imaging, maybe more? I think this would be good for the sake of organization and also compilation and linking times.
Oh well, maybe this second level is not considered at the moment, but anyway it would be a good idea to keep it in mind.
In such cases, would we prefer a nested level, or a flatter structure?

Design/CodeReorg/CodeReorg.md
```diff
@@ -0,0 +1,113 @@
...
+	|-- common
+	|   |-- algorithm: algorithm interface + impl classes (dataprocessor), property manager, basic properties
+	|   |-- constants: maths: e.g. 2*pi, physical: hbar, neutronic: neutron mass
+	|   |-- core: System.h, MantidVersion.h ...
+	|   |-- datamodel: workspaces, axes, validators, workspace property
+	|   |-- units
+	|-- geometry
+	|   |-- legacy
+	|-- io: file loader interface, file searching, fileproperty
+	|   |-- nexus: contents of our nexus directory
```
FedeMPouzols: I have the impression that I/O, which is fairly big, would benefit from further splitting. A clear-cut case could be groups of facility or format/standard specific loaders+savers, like the ILL algorithms, or the Raw-related algorithms. That would produce very cohesive and small libraries.
I'd see this as a lower priority, as there's a lot to change elsewhere, but commenting here just to keep it in the plans.

Design/CodeReorg/CodeReorg.md
martyngigg: I agree, we probably want that. It wasn't an exhaustive list yet but just to give an indicator

Design/CodeReorg/CodeReorg.md
martyngigg: :+1:

Design/CodeReorg/CodeReorg.md
OwenArnold: Anything that can acts on a `Workspace` and other meta-data and outputs a `Workspace`. `Algorithm`, `RemoteAlgorithm`  ...

Design/CodeReorg/CodeReorg.md
```diff
@@ -0,0 +1,118 @@
...
+	|-- mpi
+	|-- optimization
+	|   |-- common
+	|   |-- costfunctions
+	|   |-- minimizers
+	|   |-- models
+	|-- neutron
+	|   |-- diffraction
+	|   |-- inelastic
+	|   |-- crystallography
```
MichaelWedel: Some parts of this are quite independent from neutrons, maybe they could go into something more math-related, possibly somewhere in `common`? The parts that have something to do with neutrons (some aspects of structure factor calculation) would be appropriate to put here.

martyngigg: True. Do you think we should have a general `libmantid-algorithm-common` plus a `libmantid-neutron-common.so`?

MichaelWedel: I was thinking more along the lines of `libmantid-common-math`, which might also contain things that are currently in `Kernel`, such as `V3D` or `Matrix`.

martyngigg:
I'm closing this as we're not pursuing this at this time. I suspect the new workbench to include some of these ideas.

PR #12, opened 2016-02-10
Proposal for saving bin mask information to xml filese in the instrument view

AntonPiccardoSelg:


abuts:
Why not just make bin mask as NaN and forget, on signal, teach  Mantid to understand it and forget about any separate workflows. 

Benefit: Bin masks transparently go from workspace with instrument (where they rarely useful and used)  to analysis workspaces (where they can and are often used) and you do not need any special measures to remember abut them

PR #13, opened 2016-03-09
Spectrum Number and Workspace Index Abstraction Design Document

SimonHeybrock:


Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,72 @@
...
+#### Definitions:
+1. Workspace index = linear index starting at zero, corresponds to memory offset in vector of spectra
+2. Spectrum number = user defined index for spectra in certain workspace
+
+#### Current situation:
+
+- Conversion methods between index types (spectrum number and workspace index) are part of the workspace interface.
+- There is no consistent way of defining index ranges or lists for algorithms. This is done individually in each algorithm, and properties may be defined for spectrum number or workspace indices, and sometimes the interpretation of index properties even seems to depend on the facility.
+- Validation of index ranges and behavior on errors such as out-of-range indices is likewise done individually in all algorithms. Algorithms behave inconsistently depending on errors (throw and error, print a warning, ignore silently). There are bugs in some range validations.
+
```
OwenArnold: I might also highlight the problem of the hard-coded assumption about mapping. There are a number of areas in the codebase where the spectrum numbers are assumed to start from 1. This is usually the case at ISIS, but if it was different in the file (and it could be) it would be a nasty problem to track down.

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,72 @@
...
+- Eliminate duplicated code for validating indices and converting them to the desired type.
+- Developers should generally not have to deal with spectrum number at all. Spectrum number is important for the user interface, but irrelevant for the internals of most algorithms.
+- Developers should in the future be isolated from dealing explicitly with the local and global indices that come with an MPI implementation of Mantid.
+
+
+## User interface
+
+- The interface of algorithms should make it possible to specify spectrum number ranges and lists.
+- For better compatibility for the current interface and for users who do not like to deal with spectrum number it should also be possible to specify workspace index ranges and lists.
+
```
OwenArnold: If the user interface side of things are to be fixed. I think it would make sense to introduce a property type for spectrum ranges. At the moment either `PropertyWithValue<int>` or `ArrayProperty<int>` are used for such purposes, developers then usually have to to use a `BoundedValidator` to prevent < 0 being used. Developers are currently then responsible for specifying what the input actually means (workspace indexes, others as spectrum numbers and others as detector ids). I think that a new property type that encapsulates all of this could be a solution to the UI side of the problem.

SimonHeybrock: I agree. I had considered doing something like that in the partial implementation linked at the bottom. However, we need something more sophisticated than `ArrayProperty`. The latter translates ranges into arrays, i.e. `1-1000000` becomes a vector of 1M elements, which is inefficient, since this is just a min/max type of property.

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,72 @@
...
+  - Defines a consistent way of dealing with errors in the index specification. This is basically also part of a consistent user interface.
+  - Reduces amount of code in algorithms and makes them more readable.
+- Provide a convenience function or method for declaring all relevant algorithm properties in a single statement.
+  - Reduces amount of code in algorithms and makes them more readable.
+  - Automatically gives consistent interface.
+- Spectrum numbers can be defined differently in each workspace. As a consequence the object dealing with translation must be linked to a workspace. For example, it could be a member of the workspace.
+- We should make sure that we do not negatively influence performance when a trivial set of indices is specified by the user. Trivial can mean either the full range or a range in combination with a trivial mapping from spectrum number to workspace index (such as a simple offset of 1).
+
+A very basic example that covers some of the features described here (property declaration and range validation, but not translation from spectrum number) can be found as part of [this pull request](https://github.com/mantidproject/mantid/pull/15465).
```
OwenArnold: Looking at that PR. Not clear to me how one might handle the situation where two loop ranges over the same workspace would be handled. This situation may crop up in Reflectometry, where spectrum are summed together for different purposes. Might be worth thinking about.

SimonHeybrock: I think that should be doable. The class `IndexSet` could probably be extended to support that without too many headaches. Given that it is an abstraction, this functionality could even be added at a later point without breaking the clients (algorithms).

Design/spectrum_number_and_workspace_index_abstraction.md
stuartcampbell: Just to mention so it's documented somewhere that we actually use -1,-2 to indicate the ids of the monitors.  This was  because here they are not assigned a detector ID (which we usually just map onto a spectrum ID) so it was either that or pick a crazy big number that could never clash as the instruments add detectors.

quantumsteve:
Would an iterator-based solution meet all of these requirements?

SimonHeybrock:
@quantumsteve I think it is certainly a good idea to prove iterators (in particular to allow for using range-based for-loops), but at least for now we also need a direct index access, namely for the OpenMP `pragma omp parallel for` for-loops.

SimonHeybrock:
@quantumsteve There may also be other cases where iterators may not be sufficient or convenient, e.g., when you need neighbors of a detector/spectrum, or for grouping detectors/spectra.

Design/spectrum_number_and_workspace_index_abstraction.md
SimonHeybrock: @OwenArnold Reading your comment again, I am not sure I understood correctly when I first replied: Are you talking about two nested loops (e.g., for a cross-correlation computation), or two ranges over different parts of the workspace?

quantumsteve:
I think we want both iterators and indexes. Each interface has strengths and weaknesses.

OpenMP supports iterator-based loops since version 3.0.

The parallelism TS, recently voted into C++17, uses random access iterators. Intel has a proposal for index-based parallel loops.

SimonHeybrock:
@quantumsteve @stuartcampbell: @OwenArnold and me just had a discussion on detector IDs. Could you share your thoughts on adopting a identical or similar design also in that case, with `detectorID` (in place of `spectrumNumber`) and `detectorIndex` (in place of `workspaceIndex`).

Are there any differences that you can think of?

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -53,7 +53,7 @@ Global workspace index 0 1 2 3 4
   - Maps are build on-demand from the data stored in the vector of spectra and are returned by value.
   - In many cases the spectrum number is defined as `workspace_index + offset`, typically `offset = 1`.
     This definition will break down completely in the MPI case where instead the spectrum number would be defined as `global_workspace_index + offset`.
-    The global workspace index is, however, now immidiately available at the level where such computations are currently done.
+    The global workspace index is, however, now immediately available at the level where such computations are currently done.
```
SimonHeybrock: Thanks. Actually this should also read `not`, not `now`.

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,212 @@
...
+
+## Detailed goals
+
+- All algorithms should have a consistent interface.
+- Eliminate duplicated code for declaring properties for index ranges and list.
+- Eliminate duplicated code for validating indices and converting them to the desired type.
+- Developers should generally not have to deal with spectrum number or detector ID at all.
+  Spectrum number and detector ID are important for the user interface (or file interface), but irrelevant for the internals of most algorithms.
+- Developers must be isolated from dealing explicitly with the local and global indices that come with an MPI implementation of Mantid.
+
```
OwenArnold: Agreed on the goals. Good high-level description.

OwenArnold: Splitting into several properties (as you describe below) suggests that the inputs will be qualified to be either SpectrumNumber/GlobalIndex, or DetectorIndexes. If you haven't discussed this with @NickDraper you might want to do so, because I recall him talking about adding some kind of universal translation service, although that might have been at the UI level only (i.e. No changes to the existing `FloatArrayProperty` properties of any algorithm.

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,212 @@
...
+  Putting *global* in the UI would only create confusion.
+
+#### DetectorIDsProperty
+
+Some algorithms work directly with detectors and thus need a detector index.
+This is similar to the `SpectrumNumbersProperty` but simpler.
+Most likely it would be sufficient to provide only *detector ID* as possible input.
+
+It should be pointed out that:
+
```
OwenArnold: I think the suggestion below is sensible. Are there any algorithms that currently provide inputs that could be interpreted as any kind of id or index? Would be interested to know if breaking changes to any existing Algorithm interface is required.

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,212 @@
...
+  - Defines a consistent way of dealing with errors in the index specification.
+    This is basically also part of a consistent user interface.
+  - Reduces amount of code in algorithms and makes them more readable.
+- Provide a convenience function or method for declaring all relevant algorithm properties in a single statement.
+  - Reduces amount of code in algorithms and makes them more readable.
+  - Automatically gives consistent interface.
+- Spectrum numbers can be defined differently in each workspace.
+  As a consequence the object dealing with translation must be linked to a workspace.
+  For example, it could be a member of the workspace.
+- We should make sure that we do not negatively influence performance when a trivial set of indices is specified by the user.
```
OwenArnold: Do you have any sense as to how we can check adversely affecting performance for this? Are the performance tests system tests enough?

Design/spectrum_number_and_workspace_index_abstraction.md
SimonHeybrock: One breaking change would be naming of properties, if we want to enforce consistency.

Regarding ambiguous interpretation, there is [17073](https://github.com/mantidproject/mantid/issues/17073) (if I understand the problem correctly).

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,212 @@
...
+Algorithms then use:
+
+```python
+for index in indices:
+  ws.mutableY(index) += offset
+```
+
+There are a few more complex cases.
+For example, in `LoadEventNexus` a vector of events labeled by detector ID needs to be inserted into a workspace.
+This requires *filtering* a vector with data based on a spectrum number of detector ID.
```
OwenArnold: Do you mean you have to identify the spectrum based solely on the detector id so that the events can be added? I'm not entirely clear about this in the way that you've described it.

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,212 @@
...
+That is, we would not return a set of indices but a vector that is a subset of the input vector, i.e., the input vector with all data corresponding to other MPI ranks removed.
+
+#### Implementation and effects on existing code
+
+- Implement the `Translator` and store it in a workspace.
```
OwenArnold: 👍

Design/spectrum_number_and_workspace_index_abstraction.md
SimonHeybrock: I don't think our performance tests are in general sensitive enough for this. Probably we will be fine if we support some special cases (which are actually the most common ones):

- We want to avoid going through a `map` or `vector` for translation in the simple cases of "full-range" (all spectra), or a simple "min-max" case. As you saw, I already implemented this in the prototyping I did for the `Indexing` library. I believe this will cover 95% of cases and ensure that there will be no performance hit.

- In the other cases, when an actual list of spectra is given, I suppose that that list is commonly small (since it is often manual user input), so `map` access is negligible.

- The remaining performance critical case is loading event data. Each event has an associated detector ID, and this needs to be translated to a workspace index. This is currently done by creating a `vector` (instead of a `map`) for look-up. We certainly have to support a similar mechanism. This is a case where we will need to write some performance tests, to prevent regression.

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,212 @@
...
+  Most likely it makes sense to share this between workspaces, i.e., use `Kernel::cow_ptr`, since indices change rarely after initial creation.
+
+- We cannot provide setters and getters for spectrum numbers and detector IDs in `Ispectrum` any more.
+  Otherwise it would not be possible to maintain a consistent index map in `Translator`.
+
+  This is likely the most critical change, since it will break existing python scripts.
+  Since the fix is simple we can probably replace the existing exports by variants that print and error including a replacement suggestion, e.g., `ISpectrum::getSpectrumNo() is longer available. Replace, e.g., ws.getSpectrum(i).getSpectrumNo() with ws.getSpectrumNo(i)`.
+
```
OwenArnold: Regarding the `ws.getSpectrum(i).getSpectrumNo()` with ws.getSpectrumNo(i)` on the python side. This will be a breaking change. I believe that we are due to talk to the TSC about this. Maybe you could add it as an agenda item.

Design/spectrum_number_and_workspace_index_abstraction.md
SimonHeybrock: Yes, that appears to be how it is done. It fails if there is not 1:1 mapping from detectors to spectra, but since this is in a load routine I think it is always true.

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,212 @@
...
+- Replace or remove a series of old index mapping methods from `MatrixWorkspace`.
+
+- Implement two new property types, `SpectrumNumbersProperty` and `DetectorIDsProperty`.
+  We should also provide convenience methods to declare these properties to ensure that the properties have consistent naming across the framework.
+
+- Refactor algorithms to use the new properties.
+  This can probably done step by step and should not pose significant problems.
+
```
OwenArnold: No mention on the `Partitioning` strategy. Is that relevant for this discussion? I guess at this level it's only necessary to say that local workspace indexes will have some mapping to the global indexes.

Design/spectrum_number_and_workspace_index_abstraction.md
OwenArnold: I suggest adding performance tests well in advance for these critical cases. That way we can gather enough statistics prior to any changes.

Design/spectrum_number_and_workspace_index_abstraction.md
SimonHeybrock: Yes. Before that, we need to get a better idea of the long term goal, however: If we intend to eliminate many uses of `SpectrumNumber` and `DetectorID`, the corresponding getters and setters should eventually be used only in special cases (such as loading and saving). That is, if we now refactor to move those methods from `ISpectrum` to `MatrixWorkspace` (breaking PYthon scripts), would that mean that there will be a second breaking change later?

I have to admit I do not have a clear idea of this currently, I would be glad to hear what you think.

- Would we need a second breaking change, or would the interface stay the same, once it is moved to `MatrixWorkspace`.
- If the interface would need to change again, would it be possible to do both changes in a single step, breaking scripts only once?

Design/spectrum_number_and_workspace_index_abstraction.md
SimonHeybrock: `Partitioning` might be an implementation detail of the `Translator`, or it might be a separate entity, I don't know yet. Does it make a conceptual difference in something to be discussed here?

Design/spectrum_number_and_workspace_index_abstraction.md
OwenArnold: I suggest we break things just once. Nobody likes an API that keeps shifting under their feet. I'm not too concerned what the interface looks like. I have a slight preference for exposing the translator and therefore reducing the additions to `MatrixWorkspace`, but we have to document that well and do things that are easy/natural to casual developers.

Design/spectrum_number_and_workspace_index_abstraction.md
OwenArnold: At best, its just a footnote since the reader might be interested in what was planned, but it's not very relevant to the discussion here. All we need to know is that the `Translator` can provide such a mapping.

SimonHeybrock:
Closing this to continue discussion in new PR

PR #14, opened 2016-03-15
Histogram type

SimonHeybrock:
First draft of a design proposal for a histogram data type in Mantid. Please comment!

### DIFF
--- a/Design/histogram_type.md
+++ b/Design/histogram_type.md
@@ -188,11 +188,11 @@ The amount of new types introduced may seem appalling at first, but we usually d
     // Const accessors return by value (there is no non-const overload):
     // Counts Histogram::counts() const;
     auto counts = hist.counts();
-    // modify counts (triggers copy via cow_ptr)
+    // modify counts
     hist.setCounts(counts);
     ```
 
-  Both options involve some overhead when writing (probably less in the former case if we move the underlying `std::vector<double>` from `Frequencies` to `Counts`).
+  Both options involve some overhead when writing (in the former case we can move the underlying `std::vector<double>` from `Frequencies` to `Counts`, in the latter case we allocate memory for the temporary `Counts` object).
 
   It is not entirely clear how we would deal with the similar situation in the `BinEdges` vs. `Points` case, where a conversion is not lossless and should thus not be done automatically in the generic case.
 
--- a/Design/histogram_type.md
+++ b/Design/histogram_type.md
@@ -150,16 +150,6 @@ The amount of new types introduced may seem appalling at first, but we usually d
 5. `Variances` and `StandardDeviations` can be automatically converted, you do not have to deal with it yourself.
 
 
-#### Details
-
-- The different low-level types have been introduced in the previous section.
-  To avoid highly duplicate code and tests, such as for the four `Variance` types, these are implemented using the curiously recurring template pattern (CRTP) which provides something like "mixins".
-- Jon pointed out that it may make most sense to default to storing data as `Frequecies`, as OpenGenie did by default.
-  Many operations could probably done in this representation, and furthermore the conversion to `Counts` is cheaper, since it includes a multiplication instead of addition.
-- The current copy-on-write (COW) mechanism seems useful and should be preserved.
-  - The `typedef` `MantidVecPtr` for a `cow_ptr` is useless and confusing and should be removed.
-  - We should consider extending the COW mechanism and make sure sharing is preserved more than it is now. For example, what can we do to maintain sharing of `X` for all histograms in a workspace where `X` is modified in an identical way? The COW mechanism as it is now should break down in that case. Probably this needs to be handled at the workspace/algorithm level, and cannot be dealt with this here? Having parametric `X` could circumvent that in certain cases (see outlook).
-
 #### Interface
 
 - The interface of `Histogram` can be (partially) forwarded by `ISpectrum` and `MatrixWorkspace`, similarly to how it is implemented currently for, e.g., `ISpectrum::dataX` and `MatrixWorkspace::dataX`.
@@ -187,6 +177,7 @@ The amount of new types introduced may seem appalling at first, but we usually d
     Histogram hist;
     // Const accessors return by value (there is no non-const overload):
     // Counts Histogram::counts() const;
+    // (note that return by value does not imply copying of data: value is a cow_ptr)
     auto counts = hist.counts();
     // modify counts
     hist.setCounts(counts);
@@ -194,14 +185,114 @@ The amount of new types introduced may seem appalling at first, but we usually d
 
   Both options involve some overhead when writing (in the former case we can move the underlying `std::vector<double>` from `Frequencies` to `Counts`, in the latter case we allocate memory for the temporary `Counts` object).
 
-  It is not entirely clear how we would deal with the similar situation in the `BinEdges` vs. `Points` case, where a conversion is not lossless and should thus not be done automatically in the generic case.
+  In an initial review, several people preferred the setter style, in particular since "This seems to fit with the general approach of making things easy to use correctly and hard to misuse that should guide all APIs." (Owen).
 
-  Further thoughts and ideas for this would be really welcome!
+  It is not entirely clear how we would deal with the similar situation in the `BinEdges` vs. `Points` case, where a conversion is not lossless and should thus not be done automatically in the generic case (there is 1 free parameter in the conversion `Points` to `BinEdges`, since the latter vector is longer by 1, so this conversion is not lossless unless the bin width is constant).
+
+- Quite a few algorithms do not care whether they are, e.g., working with bin edges or points. A concrete example for this would be `ScaleX`, which just scales the values of the `X` data, regardless of whether it is dealing with histogram data or point data. As a consequence, these 'ignorant' algorithms cannot use the part of the `Histogram` interface I described so far,
+  
+  ```cpp
+  BinEdges binEdges() const;
+  void setBinEdges(const BinEdges &);
+  Points points() consts;
+  void setPoints(const Points &);
+  ```
+  
+  since the algorithm does not know and does not care what the underlying data layout is. We could ignore this ignorance and just use, e.g., `binEdges` and `setBinEdges`, but that can force a conversion of the underlying data.
+  
+  After a discussion with Michael Wedel, I would like to propose the following solution (examples are for `X`, but the same applies to `Y`):
+  
+  - Introduce an additional type `HistogramX`. This is somewhat similar to the current way of storing `X` data in an `ISpectrum`, but with very important differences.
+    The class members are:
+    ```cpp
+    cow_ptr<std::vector<double>> m_data;
+    HistogramX::XMode m_xMode; // can be XMode::BinEdges or XMode::Points
+    ```
+    That is, a *flag* *indicates* the *current* *state* *of* *the* *data*.
+    The interface is:
+    ```cpp
+    // creation and access similar to std::vector, including iterators, etc.
+    const double &operator[](size_t) const;
+    double &operator[](size_t);
+    // NO RESIZING VIA THE PUBLIC INTERFACE!
+    // Access to data via specific type. BinEdges and Points are as described before, wrappers around `cow_ptr<std::vector<double>>`
+    BinEdges binEdges() const;
+    void setBinEdges(const BinEdges &);
+    Points points() consts;
+    void setPoints(const Points &);
+    ```
+    That is, *we* *forbid* *resizing* `HistogramX`, such that client code cannot break `Histogram`.
+  
+  - `Histogram` contains `HistogramX`. The interface is now slightly extended:
+    ```cpp
+    // Access to underlying data (similar to the current `readX` and `dataX`), *but*:
+    // - state always known, allows for safe operations, length checks, ...
+    // - cannot be resized, i.e., client code cannot break the state of the Histogram
+    HistogramX &x();
+    const HistogramX &x() const;
+    const HistogramX &constX() const;
+  
+    BinEdges binEdges() const;
+    void setBinEdges(const BinEdges &);
+    Points points() consts;
+    void setPoints(const Points &);
+    ```
+
+  A usage example of this interface for `ScaleX` is:
+  ```cpp
+  Histogram histogram( /* init somehow */);
+  histogram.x() *= 0.5;
+  ```
+  
+  Maybe at this point is is worth noting that `HistogramX` has another advantage over the current `readX` and `dataX` interface. Consider, e.g., this piece of code from `BinaryOperation.cpp`:
+  ```cpp
+        // Get reference to output vectors here to break any sharing outside the
+        // function call below
+        // where the order of argument evaluation is not guaranteed (if it's L->R
+        // there would be a data race)
+        MantidVec &outY = m_out->dataY(i);
+        MantidVec &outE = m_out->dataE(i);
+        performBinaryOperation(m_lhs->readX(i), m_lhs->readY(i), m_lhs->readE(i),
+                               m_rhs->readY(rhs_wi), m_rhs->readE(rhs_wi), outY,
+                               outE);
+  ```
+  We are currently returning references to data inside a `cow_ptr`, so there is always the risk of breaking them, as explained in the code comments. In the proposed `HistogramX` we actually return a reference to a `cow_ptr`, i.e., when something triggers a copy the const references to `HistogramX` remain valid!
 
 - Not all cases for operations with histograms will be covered by the histogram type or free functions for histograms.
   Therefore we need to provide a direct interface to the underlying data as well.
 
+- As Andrei pointed out (with strong agreement form everyone), automatic rebinning should not be performed. That is:
+  - No automatic rebinning by `operator+`.
+  - We have to agree on what level of compatibility check operations should have. A good orientation point are the `BinaryOperations` like `Plus`, which seem to be checking the size of `X` but nothing else. 
+  - Operation with different sizes could later be added, but with a more explicit interface, e.g., `Histogram::rebinAndAdd(const Histogram &rhs)`.
+
 - If we want to hide things like error handling, is there a decent way of making this extensible for new operations? Are operators/functions declared as `friend` in our histogram data type? Are they members? Can we allow the use of `std::transform` on, e.g., the `Y` data? How to transform errors?
+
+
+#### Details
+
+- The different low-level types have been introduced in the previous section.
+  To avoid highly duplicate code and tests, such as for the four `Variance` types, these are implemented using the curiously recurring template pattern (CRTP) which provides something like "mixins".
+
+- Jon pointed out that it may make most sense to default to storing data as `Frequecies`, as OpenGenie did by default.
+  However, Andrei and others disagreed, based on more evidence on how this is used in the Mantid code.
+  In fact, we have 3 options:
+  
+  1. always store counts
+  2. always store frequencies
+  3. store either counts or frequencies, switch on demand or automatically
+  
+  During roll-out, we will have option (3), since as long as the old interface (`readY`, `dataY`, etc.) is used and available, we need to be able to return references to the internal data.
+  Once we have completely eliminated the legacy interface we can gather statistic, run more benchmarks, and then decide whether we switch from (3) to (1) or (2) or not.
+
+  It should be noted that due to the design of the interface the internal data structure is hidden as much as possible, and adapting things at a later points is possible.
+  In particular, this implies that we can start with implementation and roll-out before having reached a final decision on this point.
+
+  Note that the same is true for `BinEdges` vs. `Points`, as well as `Variance` vs. `StandardDeviation`.
+
+- The current copy-on-write (COW) mechanism seems useful and should be preserved.
+  - The `typedef` `MantidVecPtr` for a `cow_ptr` is useless and confusing and should be removed.
+  - We should consider extending the COW mechanism and make sure sharing is preserved more than it is now. For example, what can we do to maintain sharing of `X` for all histograms in a workspace where `X` is modified in an identical way? The COW mechanism as it is now should break down in that case. Probably this needs to be handled at the workspace/algorithm level, and cannot be dealt with this here? Having parametric `X` could circumvent that in certain cases (see outlook).
 
 
 ## Roll-out
@@ -249,3 +340,5 @@ The proposed encapsulation of histogram data opens the door for improvements tha
 - Abstraction of error handling may lead to improvements that are currently hard to implement, such as dealing with non-Gaussian errors or taking into account correlation between data.
 - If errors are just the square root of the counts the implementation may choose to not carry them explicitly to save computation time and memory.
 - Bin sizes could in some cases be defined in a parametric way, without keeping bin boundaries around explicitly. This could safe memory, memory bandwidth, and cache space.
+- Martyn suggested that underflow and overflow fields could be added, which are present, e.g., in the ISIS raw files.
+  See also here: https://thepeg.hepforge.org/doxygen/classLWH_1_1Histogram1D.html.
--- a/Design/histogram_type.md
+++ b/Design/histogram_type.md
@@ -162,7 +162,7 @@ The amount of new types introduced may seem appalling at first, but we usually d
 
 #### Interface
 
-- The interface of `Histogram` can be (partially) forwarded by `ISpectrum` and `MatrixWorkspace`, similarly to how it is implemented currently for, e.g., `ISpectrum::datX` and `MatrixWorkspace::dataX`.
+- The interface of `Histogram` can be (partially) forwarded by `ISpectrum` and `MatrixWorkspace`, similarly to how it is implemented currently for, e.g., `ISpectrum::dataX` and `MatrixWorkspace::dataX`.
 
 - The main challenge in the design of the `Histogram` interface is the way how we deal with references.
   If the internal data storage mode of, say, the Y-data is not known, it is not trivial to provide `Counts &counts()` (as a partial equivalent to the current `std::vector<double> &dataY()`), since there may be nothing to reference to.
--- a/Design/histogram_type.md
+++ b/Design/histogram_type.md
@@ -30,7 +30,7 @@ dis/count mess (ref ConvertUnits bug)
 
 The current way of dealing with histograms has various shortcomings.
 Basically we have a set of three `std::vector<double>`, for `X`, `Y`, and `E`.
-There is a single `typedef MantidVec` used for all three of them (and other data), and this is probably adding only more confusion. 
+There is a single `typedef MantidVec` used for all three of them (and other data), and this is probably adding only more confusion.
 
 The problems include:
 
@@ -64,3 +64,22 @@ The problems include:
 
 
 If errors are just the square root of the counts the implementation may choose to not carry them explicitly to save computation time and memory.
+
+Jon pointed out that it may make most sense to default to distribution data.
+
+Bin sizes could in some cases be defined in a parametric way, without keeping bin boundaries around explicitly.
+
+Do we need sub-types for `X`, `Y`, `E`, and `E-squared`? (and dist- data)?
+
+What can we do to maintain sharing of `X` for all histograms in a workspace where `X` is modified in an identical way? The COW mechanism as it is now should break down in that case. Does it need to be handled at the workspace level, or can we deal with this here? Having parametric `X` would circumvent that case in certain cases.
+
+
+## Interface
+
+At least initially not all cases for operations with histograms will be covered by the histogram type or free functions for histograms.
+Therefore we probably need to provide a direct interface to the underlying data as well.
+One option would be to have sub-types for `X`, `Y`, and `E`, let them inherit from `std::vector<double>` and just add new interface functions.
+
+If we want to hide things like error handling, is there a decent way of making this extensible for new operations? Are operators/functions declared as `friend` in our histogram data type? Are they members? Can we allow the use of `std::transform` on, e.g., the `Y` data? How to transform errors?
+
+## Roll-out
--- a/Design/histogram_type.md
+++ b/Design/histogram_type.md
@@ -18,8 +18,7 @@ There are various motivations for this:
 - Hide discrimination between histogram and distribution data.
 - Decrease risk for certain types of bugs by using the C++ type system.
 
-
-All this would remove duplicated code from the client and puts it into a central place where it can be unit-tested.
+All this would remove duplicated code from the client and put it into a central place where it can be unit-tested.
 
 
 ## Current situation
@@ -52,7 +51,7 @@ The problems include:
   - Arithmetic operations should do the right thing for any data type automatically.
 - Hide error handling.
   - For simple operations with histograms algorithm authors should not need to know how to propagate errors.
-- Enable unit testing of key concepts in Mantid:
+- Enable unit testing of key concepts in Mantid.
   - Arithmetic operations with histograms.
   - Propagation of errors.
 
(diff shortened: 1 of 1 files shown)


Design/histogram_type.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Current situation
+
+The current way of dealing with histograms has various shortcomings.
+Basically we have a set of three `std::vector<double>`, for `X`, `Y`, and `E`.
+There is a single `typedef` `MantidVec` used for all three of them (and other data), and this is probably adding only more confusion.
+
+The problems include:
+
+- No type safety:
```
OwenArnold: This is compelling, and definitely something I'd like to see more of. You might consider some meta-programming techniques to enforce type safety such as http://loki-lib.sourceforge.net/html/a00209.html

Design/histogram_type.md
```diff
@@ -0,0 +1,108 @@
...
+  - The bug in `ConvertUnits` creped in because we are currently using the same instance of a single type for both histogram data and distribution data. The actual meaning of a vector of `Y` values is basically kept as a mental flag in the brain of a Mantid developer.
+  - We are using a single type for `X`, `Y`, and `E`. It is thus easy to accidentally mix them up.
```
OwenArnold: There's also the x-uncertainty to consider. Best to talk to @picatess about this as he's been working on it for SANS

Design/histogram_type.md
```diff
@@ -0,0 +1,108 @@
...
+  - The array `E` of error values is used for both the variance and the standard deviation. The state is again carried as a flag in the brain of a developer. We even have functions such as `rebin` that return an array of `E` values that may mean either variance or standard deviation depending on a flag to the function!
+- Code duplication: Re-implementing the same thing 10 times is wasteful and error prone.
+  - Operations like adding histograms are frequent.
```
OwenArnold: Would be nice to see what examples you've found of this. I can think of a few.

Design/histogram_type.md
```diff
@@ -0,0 +1,108 @@
...
+  - Arithmetic operations automatically deal with propagation of errors.
+- Hide differences between histogram and distribution data.
+  - Internal storage type could be either, does not need to be known at the interface level?
+  - Arithmetic operations should do the right thing for any data type automatically.
+- Hide error handling.
+  - For simple operations with histograms algorithm authors should not need to know how to propagate errors.
+- Enable unit testing of key concepts in Mantid.
+  - Arithmetic operations with histograms.
+  - Propagation of errors.
+
```
OwenArnold: Enforcing the type safety should also be a goal.

Design/histogram_type.md
```diff
@@ -0,0 +1,108 @@
...
+#### Details
+
+- It is probably useful to introduce two types for vectors of standard deviation and variance, instead of the current double use of `E`.
+- Similarly we should have distinct internal types for histogram and distribution data, in case our implementation relies on this distinction. Jon pointed out that it may make most sense to default to distribution data, as OpenGenie did by default.
+- The current copy-on-write (COW) mechanism seems useful and should be preserved.
+  - The `typedef` `MantidVecPtr` for a `cow_ptr` is useless and confusing and should be removed.
+  - We should consider extending the COW mechanism and make sure sharing is preserved more than it is now. For example, what can we do to maintain sharing of `X` for all histograms in a workspace where `X` is modified in an identical way? The COW mechanism as it is now should break down in that case. Does it need to be handled at the workspace level, or can we deal with this here? Having parametric `X` could circumvent that in certain cases (see outlook).
```
OwenArnold: Good point in your example! I think the ability to share the x-array and dx-array across multiple histograms would be extremely useful.

Design/histogram_type.md
OwenArnold: Speed of both reads and writes should be preserved. Consider benchmarking the existing implementation in your roll-out steps.

Design/histogram_type.md
SimonHeybrock: I think what you linked to goes one step further than I intended at this point. My suggestion was to have, e.g., a type for the *vector* of `Y` values, *not* for the *individual* `Y` values of each bin. The latter would certainly be a good idea if we continue to have plenty of histogram-modifying code spread in the code base. However, the goal of this design is to move such code to a central point, so at this point is it not clear if the benefits of having a type for each individual value outweigh the increased complexity.

Design/histogram_type.md
SimonHeybrock: According to my list, I so far found the following:

- assignment
- set to zero
- arithmetics with other vectors, including `+`, `+=`, etc.
- arithmetics with a scalar, e.g., adding a constant to the bin-edge vector or the bin values
- comparison (`==`)
- integration
- rebin
- reverse order
- apply function to values (similar to `std::transform`)
- truncate head or tail
- get lambda (i.e., compute bin centers)

Design/histogram_type.md
SimonHeybrock: Good point. However, I think the speed for both implementations will typically be limited by memory bandwidth.

Design/histogram_type.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### Basics
+
+The details of the implementation have not been worked out yet.
+The basic concept seems simple enough:
+
+- Add one class for a histogram with an interface that provides standard operations for histogram data.
+  - Note that we do not introduce a separate type for histogram and distribution data. This distinction would be only internal, and access to either histogram or distribution data is provided via public interface methods. The goal is to eliminate this distinction from as much of the code as possible.
+- Add a few classes for `X`, `Y`, and `E` (etc.) that would mainly be used internally in the histogram class and for direct access to the underlying data. If these inherit from `std::vetor<double>`, they could also be used to keep the legacy interface for `ISpectrum` alive until the roll-out is complete.
+
```
martyngigg: I would be weary of deriving from `std::vector<double>`. See http://stackoverflow.com/questions/6806173/subclass-inherit-standard-containers/7110262#7110262

Design/histogram_type.md
martyngigg: Agreed. If we can do the detection of shared data for `X` sensibly then it does make a big difference for large instruments and is worth exploring.

Design/histogram_type.md
```diff
@@ -0,0 +1,108 @@
...
+- The current copy-on-write (COW) mechanism seems useful and should be preserved.
+  - The `typedef` `MantidVecPtr` for a `cow_ptr` is useless and confusing and should be removed.
+  - We should consider extending the COW mechanism and make sure sharing is preserved more than it is now. For example, what can we do to maintain sharing of `X` for all histograms in a workspace where `X` is modified in an identical way? The COW mechanism as it is now should break down in that case. Does it need to be handled at the workspace level, or can we deal with this here? Having parametric `X` could circumvent that in certain cases (see outlook).
+
+#### Interface
+
+- At least initially not all cases for operations with histograms will be covered by the histogram type or free functions for histograms. Therefore we probably need to provide a direct interface to the underlying data as well.
+- If we want to hide things like error handling, is there a decent way of making this extensible for new operations? Are operators/functions declared as `friend` in our histogram data type? Are they members? Can we allow the use of `std::transform` on, e.g., the `Y` data? How to transform errors?
+
```
martyngigg: The other question is whether we need more direct access to the data for things like plotting/visualization. `Qwt` has abstractions for the data source so this can probably be mitigated in those classes but it's something we need to make sure we understand.

Design/histogram_type.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Roll-out
+
+Since histograms are in widespread use in the Mantid code the introduction of the new data type is not trivial.
+However, we can probably adopt a step-by-step approach that makes the roll-out quite feasible:
+
+- The new histogram data type is typically encapsulated in an `ISpectrum` object.
+  - Currently the `X` data is part of `ISpectrum`, whereas `Y` and `E` are part of `Histogram1D`. We somehow have to change this, or provide the option to have a histogram without `Y` and `E` data. This may be the main difficulty of these first steps.
+  - The interface of `ISpectrum` could be kept unchanged at first. The data access functions would then forward to the internal `X`, `Y`, and `E` data inside the new histogram type.
```
martyngigg: I think this is another classic case where `ISpectrum` should be a pure interface and have no implementation of its own. `EventList` & `Histogram1D` can still share some portion of implementation for X if it's in a separate class of its own. Maybe that's a first step?

Design/histogram_type.md
SimonHeybrock: @martyngigg I agree that is not ideal (in fact I read that exact stackoverflow comment before writing this design document).

I think there are a few things to consider when taking a decision:

1. Compatibility with the current interface: This is mainly a roll-out issue, but it may take a long time until it is complete, since `X`, `Y`, and `E` are used everywhere. Can we achieve the same thing if we let, e.g, `readX` return a reference to a `std::vector<double>` *member* of a new histogram type?
2. We do not need polymorphism. When inheriting that would still leave the risk of deleting via a base class pointer. Do we generally avoid that?
3. If we inherit from `std::vector`, we would still implement most functionality via algorithms.
4. Would inheriting from `std::vector` vialotate the Liskov substitution principle? I am tempted to say so, which would be good, but I am not entirely sure.
5. `std::vector` has quite a big and nice interface (e.g., for initialization), which is used frequently. If we have the vector as a member we need to replicate that for each of our new types, which adds a lot of code.
6. Performance: If we chose composition have have one extra indirection for data access. If we inline those access methods (e.g., `operator[]`) there should be little to no negative performance impact.

I think writing this clarified a few things also for myself. If we can find a nice way of dealing with 1.) the only remaining reason for inheritance is 5.). 5.) is a rather 'local' problem, so it is dealt with easily with some extra work.

martyngigg: I see your points in that it's mainly a compatibility issue and we certainly don't want polymorphism here. I think your suggestion of returning the reference to `member` should achieve the same thing.

While I agree that the interface for vector is nice, does all of it make sense for the new types? Currently any of the length-modifying operations are allowable but not necessarily a good idea in most places given the assumptions that  all histograms in a workspace are the same length.

Design/histogram_type.md
SimonHeybrock: @OwenArnold @martyngigg: I can not see a way to deal with this on the histogram-level. We can however put this logics into `SpectrumAlgorithm::for_each`.

Do you have other ideas?

OwenArnold: @SimonHeybrock: If the `Histogram` takes the `x-array` as a COW pointer in it's constructor then it should be straight forward to achieve the sharing. 

Other thoughts:
* It would also be very useful to useful to share the `x-array` across workspaces for the multi period scenario.
* Could we do something similar for the `y-array`, `e-array`? The majority of workspaces are sparse, so maybe a shared zero_counts array, would be useful.

martyngigg: @SimonHeybrock I'm not sure we can deal with at the histogram level. Generally it really needs some knowledge of the operation and the current state to know whether you can keep sharing things afterwards. doesn't it? For example `ConvertUnits` to an angle-dependent unit cannot share X afterwards but using a unit that is essentially a scale factor can keep the shared status.

SimonHeybrock: @OwenArnold What you describes works for sharing initially. However, the issue here is the following: Say we have a workspace where `X` is the same for all spectra, this is shared (also in the current implementation). If you now add an offset to all `X` the sharing is broken, since you modify each individual `X` (and thus trigger the copy-on-write).

Regarding your other thoughts:
- Sharing across workspaces: Not sure I understand what you are referring to, is this different from the current case?
- Sharing is `Y` and `E`: This is also done now. The only current shortcoming is that the only way to copy `Y` and `E` while maintaining sharing is currently via the copy-constructor. But in principle it is done and happens in certain algorithms.

SimonHeybrock: @martyngigg I think it could be put into *certain* `for_each` implementations. Your angle dependence is a good example: There, the functor passed to `for_each` takes an extra argument (the angle), which makes sharing impossible. Basically, we can share automatically in any `for_each` call thus takes a `const` functor that takes only `dataX` as an argument (and nothing else). For example, the case with a constant shift of `X` would work (the shift would be stored in the functor).

OwenArnold:
I think the design could use a class diagram to help understand what's going to be put in place. Either a screenshot of a whiteboard or a graphviz to png export like [this](https://github.com/mantidproject/documents/blob/master/Design/Instrument-2.0/GeometryClasses.png) would be useful.

SimonHeybrock:
@OwenArnold Thanks! I will have a look.

Following a discussion with Owen last week, it may also be necessary or advisable to add support for "point data" to the histogram type. I am currently trying to figure out details/requirements and would be happy about comments.

OwenArnold:
@SimonHeybrock. I think you'd almost certainly want to encapsulate the point data in the same way. Does Histogram stop being the right name for the type if we're dealing with point data?

SimonHeybrock:
@OwenArnold Not sure about the name. Are there alternatives?

SimonHeybrock:
@OwenArnold 
My attempt a a class diagram: [histogram_type_class_diagram.pdf](https://github.com/mantidproject/documents/files/201957/histogram_type_class_diagram.pdf)

The implementation of the lower level bits seems straight forward . I am using the curiously recurring template pattern (CRTP) to build something like 'mixins' (omitted most inheritance arrows for clarity). This avoids code duplication in the classes `BinWidths`, `Bin...`, ..., and also in their tests.

The main class `Histogram` merges all the lower level bits together. The internal representation of data is not fixed, e.g., errors can be stored 'implicitly' (computed as sqrt of counts), 'variance', or 'standard-deviation' and data could be the default 'values' (equiv. to the current 'distribution data') or 'counts'.

OwenArnold:
@SimonHeybrock Regarding names. I don't have a better one to hand. I'm just aware that as soon as you're dealing with point data, calling it a Histogram might be misleading. This is a minor point. Renaming can happen later if needed.

OwenArnold:
@SimonHeybrock Regarding the `Histogram` on you class diagram. You might want to add r-value reference setters in addition to the l-value reference versions.

Design/histogram_type.md
```diff
@@ -0,0 +1,108 @@
...
+The basic concept seems simple enough:
+
+- Add one class for a histogram with an interface that provides standard operations for histogram data.
+  - Note that we do not introduce a separate type for histogram and distribution data. This distinction would be only internal, and access to either histogram or distribution data is provided via public interface methods. The goal is to eliminate this distinction from as much of the code as possible.
+- Add a few classes for `X`, `Y`, and `E` (etc.) that would mainly be used internally in the histogram class and for direct access to the underlying data. If these inherit from `std::vetor<double>`, they could also be used to keep the legacy interface for `ISpectrum` alive until the roll-out is complete.
+
+#### Details
+
+- It is probably useful to introduce two types for vectors of standard deviation and variance, instead of the current double use of `E`.
+- Similarly we should have distinct internal types for histogram and distribution data, in case our implementation relies on this distinction. Jon pointed out that it may make most sense to default to distribution data, as OpenGenie did by default.
```
NickDraper: I'm a little concerned about specifying id the data is a distribution or histogram at this level.  While I agree that it makes sense computationally, I cannot come up with a valid use case where you would have a mixture within a single workspace.  Taking this approach would make that possible, and if it is possible within the code someone somewhere will do it, and I suspect that very few algorithms would handle that situation gracefully (and they shouldn't have to).  The current approach of having the distribution flag on the workspace has it's flaws, but that is at the level it makes sense to define it for a workspace, either all distribution or none.

NickDraper: Also we would need to be able to force the data from one to the other (with and without performing the division by bin width.

Certainly in some past scripts and probably still current there were situations where the distribution flag had to be set manually in python because some combination of other operations had converted the data into a distribution without setting the flag.

SimonHeybrock: Having a mixture in a single workspace is also possible right now, so this is not an issue that this design introduces.

On the contrary: In the current design it is technically *impossible* to verify whether data is histogram or distribution (and it is not hard to find a scenario where this could happen currently). In the proposed design this verification is possible and could be done at the workspace level. This would be part of a design change of, e.g., `MatrixWorkspace`.

SimonHeybrock: Having 'some combination of other operations had converted the data into a distribution without setting the flag' is exactly what we need to avoid. It leads to bugs, for example in `ConvertUnits`. Forcing is of course always possible, but it should not be made easy since it is really dangerous. You can still use `histogram.setFrequencies(Frequencies(histogram.getCounts().data()))`.

Design/histogram_type.md
```diff
@@ -34,9 +34,21 @@ The problems include:
...
+    - arithmetics with other vectors, including `+`, `+=`, etc.
+    - arithmetics with a scalar, e.g., adding a constant to the bin-edge vector or the bin values
+    - comparison (`==`)
+    - integration
+    - rebinning
+    - reverse order (maybe just part of other operations)
+    - apply function to values (similar to `std::transform`)
+    - truncate head or tail
+    - get lambda (i.e., compute bin centers)
+- Dealing with different cases of data representation. For example, many algorithms require point data. Currently we have manual checks for `isHistogramData` and subsequent conversions to point data scattered throughout our code.
```
OwenArnold: I  think these justifications are compelling reasons to encapsulate this.

Design/histogram_type.md
```diff
@@ -99,6 +236,12 @@ Furthermore, we will be able to decide whether or not we provide direct access t
 This direct access would not be identical to the current one, since it would rely on the new types for these data structures.
 
 
+## Python interface
+
+- There should be no major difficulties providing a Python interface.
```
OwenArnold: One thing I've just thought about is backwards-compatibility in the python interface. we currently expose numpy arrays via `readX()`. We're going to have to keep these around until we can deprecate their use. We should still introduce and expose the `Histogram` type an maybe give some talks and demos on why this should be the preferred method of interaction.

Design/histogram_type.md
```diff
@@ -0,0 +1,251 @@
...
+    // Const accessors return by value:
+    // const Counts Histogram::constCounts() const;
+    const auto &counts = hist.constCounts();
+    // Non-const accessors return by reference and converts internal storage mode:
+    // Counts &Histogram::counts();
+    auto &writableCounts = hist.counts();
+    // modify writableCounts
+    ```
+
+  - Alternatively, we could eliminate the option of getting non-const references, and instead do everything via setters, e.g.,
```
OwenArnold: I prefer the setter approach in this case. More work to write what you want and less work to get the read-only counts. There may be a danger in the above that someone wants the read-only counts, but accesses it in an immutable way.

Design/histogram_type.md
```diff
@@ -0,0 +1,251 @@
...
+    // Const accessors return by value (there is no non-const overload):
+    // Counts Histogram::counts() const;
+    auto counts = hist.counts();
+    // modify counts
+    hist.setCounts(counts);
+    ```
+
+  Both options involve some overhead when writing (in the former case we can move the underlying `std::vector<double>` from `Frequencies` to `Counts`, in the latter case we allocate memory for the temporary `Counts` object).
+
+  It is not entirely clear how we would deal with the similar situation in the `BinEdges` vs. `Points` case, where a conversion is not lossless and should thus not be done automatically in the generic case.
```
OwenArnold: Can you clarify what you mean by "not lossless"?

Design/histogram_type.md
```diff
@@ -0,0 +1,251 @@
...
+
+  Further thoughts and ideas for this would be really welcome!
+
+- Not all cases for operations with histograms will be covered by the histogram type or free functions for histograms.
+  Therefore we need to provide a direct interface to the underlying data as well.
+
+- If we want to hide things like error handling, is there a decent way of making this extensible for new operations? Are operators/functions declared as `friend` in our histogram data type? Are they members? Can we allow the use of `std::transform` on, e.g., the `Y` data? How to transform errors?
```
OwenArnold: I think the use of `friend` can be sensibly limited to allowing non-member binary functions to access internals. We should expose non-member binary functions where possible.

Design/histogram_type.md
SimonHeybrock: Well, the point data vector is 1 shorter than the bin edge data. When Mantid converts PointData to HistogramData it uses the midpoint between two points. For constant bin sizes that is correct, in all other cases your bin edges or points would slowly drift if you keep converting back and forth.

OwenArnold: Ok. That's what I hoped that you meant over accumulated errors from floating point operations. That being the case, and given that you do not want to allow these operations to occur by default. I would give even more credit to the setter approach where write access can be better controlled. i.e. don't provide the non-const T& access at all.

SimonHeybrock: No, it is not just accumulated floating point errors! The errors are much much larger. The problem is that there is no unique mapping from points to bin edges (basically you have an under-determined set of equations, there is 1 free parameter).

Design/histogram_type.md
martyngigg: I too prefer the setter style as it more explicit. In the second style, what's the reason that "const accessors return by value" and not by const reference? Copying every time for read operation seems like overkill.

martyngigg:
Is there any appetite for storing underflow and overflow fields?

Design/histogram_type.md
SimonHeybrock: @martyngigg The internal data representation is not necessary equal to what the getter access, e.g., data could be stored as `Frequencies`. We then cannot return a reference to `Counts` (unless we would want to change the internal representation on read).
When `Frequencies` is accessed we could either return a reference (if we fix the representation, but the interface would be non-uniform), or also return by value. In that case there is no actual copy of the data, since `Frequencies` is a wrapper around `cow_ptr<std::vector<double>>`.

SimonHeybrock:
@martyngigg Could you elaborate on what 'underflow and overflow fields' are?

martyngigg:
@SimonHeybrock I've only ever encountered them in high energy physics where the simplest case just carries two extra "counters" for the cases where a value is requested to be added to the histogram outside of the defined range: https://thepeg.hepforge.org/doxygen/classLWH_1_1Histogram1D.html

SimonHeybrock:
Sounds useful. I think that is something that can easily be added later without breaking things.

martyngigg:
Agreed. It would be simple extension.

Design/histogram_type.md
```diff
@@ -0,0 +1,251 @@
...
+
+We introduce types for all the different interpretations of a `std::vector<double>` in this context.
+The new types would be simple wrappers around `std::vector<double>` (or rather `Kernel::cow_ptr<std::vector<double>>`), and add a suitable interface to it.
+In particular:
+
+1. X-data: We need to cover two cases and introduce two types: `BinEdges` and `Points`.
+2. Y-data: We need to cover two cases and introduce two types: `Counts` and `Frequencies`.
+3. Errors: All four types introduced by 1. and 2. can come with uncertainties.
+   They can be represented as variances or standard deviations.
+   We thus introduce 8 types: `BinEdgeVariances`, `BinEdgeStandardDeviations`, `PointVariances`, `PointStandardDeviations`, `CountVariances`, `CountStandardDeviation`, `FrequencyVariances`, `FrequencyStandardDeviations`.
```
AndreiSavici: Not sure I ever saw variances or standard deviations for `BinEdges`. For `Points` is OK

Design/histogram_type.md
```diff
@@ -0,0 +1,251 @@
...
+   ```cpp
+   Histogram hist1;
+   Histogram hist2;
+   // Initialize histograms
+   // ...
+   // Adding histograms can automatically add `Y`, and propagates the errors properly
+   auto result = hist1 + hist2;
+   ```
+
+   Details of the `Histogram` interface are still open, but it should be simple to extend it once the roll-out is done.
```
AndreiSavici: I think the interface should be discussed and decided before roll-out. I am especially concerned about the cases where the `X` arrays are different.

Design/histogram_type.md
```diff
@@ -0,0 +1,251 @@
...
+   As long as the standard deviation is simply the square root of the counts it will be computed and propagated under the hood.
+
+5. `Variances` and `StandardDeviations` can be automatically converted, you do not have to deal with it yourself.
+
+
+#### Details
+
+- The different low-level types have been introduced in the previous section.
+  To avoid highly duplicate code and tests, such as for the four `Variance` types, these are implemented using the curiously recurring template pattern (CRTP) which provides something like "mixins".
+- Jon pointed out that it may make most sense to default to storing data as `Frequecies`, as OpenGenie did by default.
```
AndreiSavici: I disagree. How many times are we interested in data as frequencies? And since you noted that this histogram object might be a more general case than TOF neutron scattering, how do we store frequencies for quantities where frequencies do not make sense? For example, I want to store counts as a function of temperature

Design/histogram_type.md
SimonHeybrock: Currently we have `Dx` as an equivalent to `X`, i.e., we support both. Since we support conversion between `BinEdges` and `Points` we also have to support conversions between their respective uncertainties.

Design/histogram_type.md
SimonHeybrock: @AndreiSavici Good point, I agree.

My current opinion on this is:
- No automatic rebinning by `operator+`.
- We have to agree on what level of compatibility check operations should have. Maybe a good orientation point are the `BinaryOperations` like `Plus`, which seem to be checking the size of `X` but nothing else?
- Operation with different sizes could later be added, but with a more explicit interface, e.g., `Histogram::rebinAndAdd(const Histogram &rhs)`.

@OwenArnold @martyngigg What are your views of this matter?

Design/histogram_type.md
SimonHeybrock: @AndreiSavici I used the term 'frequency' in a more general interpretation here, it could be, e.g., spatial frequency, angular frequency, or in you example 'thermal frequency' (counts/Kelvin). So basically `Frequency` would always refer to `Counts` divided by `BinWidth`, and the interpretation of `Frequency` would follow from the meaning of the `BinWidth`, i.e., the `X` unit.

I am open to renaming `Frequency` if you think in can cause trouble. @OwenArnold @martyngigg, what do you think?

SimonHeybrock: @AndreiSavici Regarding "How many times are we interested in data as frequencies?": I don't know, how many times are we interested in data as counts? Or rather how many times do we need to work with (apply algorithms to) data as counts, vs. how many times as frequencies?

It would help me a lot if you could elaborate a bit on the numbers and give some examples.

Design/histogram_type.md
OwenArnold: I also think that automatic rebinning should not be performed. I'd be cautious about doing anything automatic or unexpected via the binary operators. Explicit convenience methods to rebin and add would be fine in addition to `operator+`.

AndreiSavici:
@SimonHeybrock I know what you mean by frequency. And my point is that in the workflow that we have to reduce diffuse scattering or single Crystal inelastic scattering we NEVER use distribution. The only time when I need distribution is when I look at a single spectrum on a TOF instrument

SimonHeybrock:
@AndreiSavici Ok, so you are saying storing data as distribution could be reducing performance in that case? That might be, do you have a link to the script, then I can have a look.

AndreiSavici:
@SimonHeybrock To see some examples just look at the usage examples for MDNormSCD and MDNormDirectSC. Also look at the workflow of DGSReduction, where we do several convert units. And I am not as much worried about performance, but, if I am a new programmer starting to work on Mantid, the decision to store as frequencies does not make sense to me. The document implies that it's the logical choice, and I do not understand why.

SimonHeybrock:
@AndreiSavici The decision will be invisible, this is hidden inside the `Histogram`. The interface will look similar to this:

```cpp
Counts counts();
void setCounts(const Counts &counts);
Frequencies frequencies();
void setFrequencies(const Frequencies &frequencies);
```

This is one of the main design points: Get rid of the distinction and conversions that we currently have to do on the client side. You just use whatever you prefer.

SimonHeybrock:
Regarding `ConvertUnits`: Most of the conversions between distribution and histogram data in there are not necessary, I think. This should be implicitly fixed when we complete the roll out the `Histogram` type.

SimonHeybrock:
After a discussion with @AndreiSavici I want to clarify how data can be stored internally in a `Histogram`.

As I wrote in my document, Jon's opinion was that the way OpenGenie did it (always storing data as a distribution may make most sense). In Andrei's opinion the opposite is true.
To be clear: At this point no decision has been made, and I am not strongly suggesting a particular option. In fact, we have 3 options:

1. always store counts
2. always store frequencies
3. store either counts or frequencies, switch on demand or automatically

During roll-out, we will have option (3), since as long as the old interface (`readY`, `dataY`, etc.) is used and available, we need to be able to return references to the internal data.
Once we have completely eliminated the legacy interface we can gather statistic, run more benchmarks, and then decide whether we switch from (3) to (1) or (2) or not.

Design/histogram_type.md
yxqd: I agree with Andrei that "frequency" is confusing. I would recommend
"density" instead of "frequency".

Dr. Jiao Lin
Computational Scientist
Oak Ridge National Lab

yxqd: Also, I think counts should be kept in the memory, while density would be
just calculated on the fly when requested.

abuts: as histograms are widely used with experiment, when x-axis is time, density is also confusing as refers to number/volume. As we have number/time I would suggest **Intensity**. 
Is this what is currently displayed with proper choice of the normalisation on Mantid plots?

SimonHeybrock: @abuts We do not, in general, have number/time either, after the first `ConvertUnits`, I think?

abuts: yep, considering all various usage scenario, I would go with frequency.

abuts:
I would also avoid setters in favour of non-constant accessors. Its ok in C++ but these operations are exposed to python, and there is already tons of custom code, which uses access mechanism. 
Better to have common interface for C and Python.

SimonHeybrock:
There is one aspect of the data access that has not yet been discussed sufficiently: Quite a few algorithms do not care whether they are, e.g., working with bin edges or points. A concrete example for this would be `ScaleX`, which just scales the values of the `X` data, regardless of whether it is dealing with histogram data or point data. As a consequence, these 'ignorant' algorithms cannot use the part of the `Histogram` interface I described so far,

```cpp
BinEdges binEdges() const;
void setBinEdges(const BinEdges &);
Points points() consts;
void setPoints(const Points &);
```

since the algorithm does not know and does not care what the underlying data layout is. We could ignore this ignorance and just use, e.g., `binEdges` and `setBinEdges`, but that can force a conversion of the underlying data.

After a discussion with @MichaelWedel, I would like to propose the following solution (examples are for `X`, but the same applies to `Y`):

1. Introduce an additional type `HistogramX`. This is somewhat similar to the current way of storing `X` data in an `ISpectrum`, but with very important differences.
  The class members are:
  ```cpp
  cow_ptr<std::vector<double>> m_data;
  HistogramX::XMode m_xMode; // can be XMode::BinEdges or XMode::Points
  ```
  That is, a *flag* *indicates* the *current* *state* *of* *the* *data*.
  The interface is:
  ```cpp
  // creation and access similar to std::vector, including iterators, etc.
  const double &operator[](size_t) const;
  double &operator[](size_t);
  // No resizing via the public interface!
  // Access to data via specific type. BinEdges and Points are as described before, wrappers around `cow_ptr<std::vector<double>>`
  BinEdges binEdges() const;
  void setBinEdges(const BinEdges &);
  Points points() consts;
  void setPoints(const Points &);
  ```
  That is, *we* *forbid* *resizing* `HistogramX`, such that client code cannot break `Histogram`.

2. `Histogram` contains `HistogramX`. The interface is now slightly extended:
  ```cpp
  // Access to underlying data (similar to the current `readX` and `dataX`), *but*:
  // - state always known, allows for safe operations, length checks, ...
  // - cannot be resized, i.e., client code cannot break the state of the Histogram
  HistogramX &x();
  const HistogramX &x() const;
  const HistogramX &constX() const;

  BinEdges binEdges() const;
  void setBinEdges(const BinEdges &);
  Points points() consts;
  void setPoints(const Points &);
  ```
   
Comments are welcome!

OwenArnold:
@SimonHeybrock I agree, for such algorithms the user does not care about the X representation. 

@SimonHeybrock, so with `HistogramX` how would the client perform a `ScaleX` type operation. I'd be interested to see what a code snippet for that looks like.

SimonHeybrock:
@OwenArnold Well, it depends on whether the current behaviour of `ScaleX` is a bug or a feature (currently it does not rescale `Y` of distribution data, i.e., `ConvertToDistribution` and `ScaleX` do not commute). Assuming it is a feature, we can write:

```cpp
Histogram histogram( /* init somehow */);
histogram.x() *= 0.5;
```

OwenArnold:
@SimonHeybrock. Indeed, we might have to handle the fact that this is a feature. I guess people may use this to massage their x-data into the right format after loading it.

AndreiSavici:
Ready to merge. :shipit:

PR #15, opened 2016-03-22
Generic Shape Absorption Corrections

martyngigg:
This design describes the intended approach to being able to correct data for the affects of absorption on a sample of any shape along with a container


NickDraper:
I've had a look at this and discussed it with Martyn.

Notes:

1. I have a preference for using the extended XML as we already have support for much of this at the moment, but the XML needs to be hidden from the top level UI.
1. We probably need to create an XML high level object of can as that will be used a lot.
1. Sample containers need to also have some concept of the sample volume (inside of a can for a can). This needs to be able to support at least 0 and 1 variable dimension, height for a powder in a can.
1. The python layer to help with writing XML might be useful in some cases, but should be considered a tool for use by developers in creating the definitions, and developed as such.
1. We need to consider how the library of components should work.  It should be filterable by facility and instrument, but many components are shared across instruments.
1. If we have one component per file then we will have two names (the file name and type_name).  only one should be needed and unique.

DanNixon:
Assuming that a custom parser would have to be written for an SCAD-like language I see no reason why you could not introduce syntax that would allow children of a tree node to have a material (much like the colour/texture of sub trees can be changed in the render generated n OpenSCAD).

For instance, the vanadium can example could become this:
```
module VanCan()
{
  mantid_material(ChemicalFormula="V", NumberDensity=0.072)
    cylinder(d=8, h=40);
}
```

To allow users to render this in OpenSCAD you could provide an SCAD library that is shipped with Mantid that replaces the names of commands such as `material` in the above example with modules that just render all children, i.e.
```
module mantid_material()
{
  children();
}
```

peterfpeterson:
I would like to re-empasize the importance of keeping the calculation of the correction separate from applying it. Two big reasons:

1. We can save the correction (to disk?) and apply it on a whole range of runs in a parametric series (e.g. temperature).
2. People will want to overly/compare the results with existing corrections from existing software (e.g. gudrun).

Otherwise, the big issue I have is that we currently store the `Material` as an attribute on the `Run` itself. Since it is only accessed by a couple of algorithms we can move the information how we want. There should be another usage example of a sample, in  a vanadium can, in a heat shield. This is so common that omitting it would be foolish.

We also still need to do a related re-design of the `Material` class itself to hold information about the [wavelength dependence of the scattering information](https://www-nds.iaea.org/ngatlas2/).

martyngigg:
@peterfpeterson Might you have a link to some type of schematic of that type of set up so I can get a better idea of what it looks like?

peterfpeterson:
Here is a cartoon.

martyngigg:
Thanks, @peterfpeterson. Just so I'm clear about the details: there is a cylindrical Vanadium can, with a given wall thickness. The sample is a powder that fills part of or all of the red-hashed volume. The heat shields are then thin, cylindrical sheets or a known thickness?

peterfpeterson:
The thickness should be known. Generally the inner is a solid piece of metal and the outer is a foil-type thing that isn't quite cylindrical.

peterfpeterson:
Technical drawing of a "real" one file:///home/pf9/Downloads/50mm%20Orange%20Drawing%20CR1-16085%20%20CR3-0173.pdf

martyngigg:
What bits are in the beam?

stuartcampbell:
Don't forget the case for annular samples as well.  These basically will have a 'whole' in the Centre of the sample, so an additional concentric circle that is the inner circumference of the can.

martyngigg:
Thanks, @stuartcampbell. This will be possible by just changing the `<sample-geometry>` to `<annulus>` rather than cylinder. What I haven't shown in the example is that any of the geometries can be defined to be some CSG object so in theory any shape is possible.

Perhaps what I have also not made clear enough is the link between the arguments provided to the user `SetSample` command and how they affect the setup. I will improve that now.

Design/AbsorptionCorrections/GenericShapeAbsorptionCorrection.md
```diff
@@ -0,0 +1,249 @@
...
+* support for inelastic data
+
+Users will be given a new algorithm called `CalculateSampleCorrection` that will form the interface to running different correction algorithms. These algorithms will assume
+that the metadata required, such as geometry and material properties is all completely defined on the workspace. An example of running the algorithm would be:
+
+```python
+# assume w1 has metadata already set
+w1 = ConvertUnits(w1, Target="Wavelength")
+w1_abs = CalculateSampleCorrection(w1, Method="MonteCarlo",
+    MethodArgs={"NLambda": 500, "NEvents": 300})
```
OwenArnold: What property do you intend to use for `MethodArgs`?

martyngigg: My intention was to create a new Property type that stores a `PropertyManager` on the C++ side. A user would simply use the more familiar Python `dict` that would be transformed to a C++ PropertyManager for the algorithm to use. This will allow us to support mixed value-types in the inputs rather than going through strings.

My main aim here was to avoid the usual flood of properties that a top level algorithm like this usual acquires that make it cumbersome to use. This way the interface stays more stable and it gives us flexibility to remap the arguments underneath as well if we need to.

FedeMPouzols: I think that this new property with `dict` interface would be incredibly useful not only for this design but in many places where the code would be nicer and more Pythonic. How would you handle the types of the dictionary entries? Using existing property types?

martyngigg: Yes. It should be able to handle any property type that we currently use. The only limitation would be that the keys would always have to be strings.

peterfpeterson:
[DataProcessorAlgorithm](https://github.com/mantidproject/mantid/blob/master/Framework/API/src/DataProcessorAlgorithm.cpp) already has a lot of what you are talking about with `dict` of other variables. There is a separate [PropertyManagerDataService](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/PropertyManagerDataService.h) that holds them. The only real problem is that they are passed by name, but the pointers are automatically exposed as `dict` and have all the bits of type conversion that you'd expect.

martyngigg:
True. I'd like to be able to avoid the name thing at all and just have users interact with the `dict` if possible. Hopefully it is just a small layer on top of what is already there.

PR #16, opened 2016-03-23
Requirements from meeting with scientists 23/3/16

tom-perkins:
Requirements for MuonAnalysis interface concerning simultaneous fitting functionality.
This centered around discussing some mockups.

### DIFF
--- a/Design/Muon/MuonAnalysis_sequential_fitting.md
+++ b/Design/Muon/MuonAnalysis_sequential_fitting.md
@@ -1,0 +1,47 @@
+### Meeting about sequential fitting in MuonAnalysis ###
+#### 23/3/16 ####
+
+Discussed mockups available at `Design/Mockups/muon/MuonAnalysis.bmpr` - the _Data Analysis_ tab
+
+This covers adding functionality for simultaneous fitting, which is Roadmap issue [#15518](https://github.com/mantidproject/mantid/issues/15518)
+
+Comments from scientists:
+
+1. **Selection of what data to fit**  
+  (bottom right of the mockup)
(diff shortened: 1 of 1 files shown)


tom-perkins:
This is the mockup for the 'selection of what data to fit' part of the document:
data-selection

and this is the mockup of the parameters for the fit function:
parameters

tom-perkins:
Attaching the whole mockup for this tab.
[DataAnalysisTab.pdf](https://github.com/mantidproject/documents/files/186235/DataAnalysisTab.pdf)

tom-perkins:
@raquelalvarezbanos @Anders-Markvardsen Tagging you so you can see the results of the meeting

tom-perkins:
This is proposed to replace the "Data" part of the muon fit property browser, based on comments from the meeting.
It will be a reusable Widget with its own tests.
dataselector

tom-perkins:
Suggest adding radio buttons like this:
radiobuttons
On the front tab, `99990-2, 99994` means these runs will be co-added. But in the mockups discussed so far, it is implied these runs will be fitted simultaneously. We need to offer both of these use cases, but differentiate between them.
*To discuss:* other ways of achieving this?

tom-perkins:
All required features for the 'Parameters' section are already in the control from the existing multi-dataset fitting interface, which is in MantidWidgets. It has the ability to set/fix different initial values for different datasets.
paramwidget

tom-perkins:
This was merged into Mantid master and will be in release 3.8. See https://github.com/mantidproject/mantid/pull/16855

PR #17, opened 2016-03-23
Instrument 2.0 design notes

OwenArnold:
This is not a fully-fledged design document, so is exempt from the usual TSC rules. Specifically, any development work will be proved against the prototype first. The development will not go directly into Mantid.

I've created this as a PR to facility further collaboration and discussion regarding high-level facades required to bind the `Workspace` to the `InstrumentTree`.


Design/Instrument-2.0/design-notes.md
```diff
@@ -0,0 +1,89 @@
...
+  - To facilitate the free data layout, the interface should be `SpectrumInfo::twoTheta(size_t index)` and *not* along the lines of the current implementation with `Workspace::getDteector(size_t index)::twoTheta()`.
+  - Accessors should probably be declared inline for performance reasons.
+
+### Write access
+
+Write access requires modification of all three data structures, `InstrumentTree`, `DetectorInfo`, and `SpectrumInfo`.
+I (Simon) do not yet have an entirely clear picture of this.
+Notes:
+
+- Do we always go via `InstrumentTree` upon modification? For example, masking a spectrum or a detector could be handled at the `SpectrumInfo` or `DetectorInfo` level. These components could also have a copy-on-write mechanism. We would then still share the same `InstrumentTree` for workspaces that have different masking.
```
OwenArnold: @SimonHeybrock Interesting point. I'm of the opinion that the masking relates more to the data than the fundamental geometry. I wouldn't see maksing as a property of a geometric detector. So on that basis I think that modification would go directly against the `SpectrumInfo` `DetectorInfo`.

Design/Instrument-2.0/design-notes.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Structure
+
+Algorithms that work with workspaces need to access information about spectra.
+Therefore:
+
+- `SpectrumInfo` should be a member of workspace.
+- `SpectrumInfo` owns `DetectorInfo`.
+- `DetectorInfo` owns `InstrumentTree`.
+
```
OwenArnold: I think an important related point is that the `DetectorInfo` should be constructible only with an `InstrumentTree_uptr` as an argument.
* `DetectorInfo` doesn't make much sense without an `InstrumentTree`
* We want IOC, so the contents of the `DetectorInfo` should be injectable as far as possible
* The `InstrumentTree` is freely copied. So it's a reasonable assumption that we're going to make our `DetectorInfo` from an existing pre-loaded instrument. At the `Workspace` level, this means that we copy a `Workspace` and we also get a smart copy of our `InstrumentTree` too. I think that is the right semantics.

Design/Instrument-2.0/design-notes.md
```diff
@@ -0,0 +1,89 @@
...
+If no detectors are grouped, there is a 1:1 mapping between spectrum and detector.
+`SpectrumInfo` and `DetectorInfo` should thus share data, to avoid waste of memory and CPU resources.
+This can be facilitated by:
+
+- `GeometryDataArray` is used to store vectors of data, such as L2, twoTheta, position, masking flags, monitor flags, ....
+- `SpectrumInfo` and `DetectorInfo` have a field `std::shared_ptr<GeometryDataArray>`. By default both point to the same instance. If detectors are grouped, `SpectrumInfo` allocates its own `GeometryDataArray`.
+
+### Interface
+
+The interface to `SpectrumInfo` and `DetectorInfo` would be almost identical.
```
OwenArnold: @SimonHeybrock Any thoughts on how you might prefer to see this realised in the class structure. Most likely we want to have a base-class for these to prevent duplication. However only the `DetectorInfo` should have an `InstrumentTree`. The `DetectorInfo` should be optional somehow (maybe boost::optional) on the `SpectrumInfo`.

Design/Instrument-2.0/design-notes.md
```diff
@@ -0,0 +1,89 @@
...
+Furthermore, we need to provide access to lower-level data, e.g., via a method `const Detector &DetectorInfo::detector(size_t index)`.
+For `SpectrumInfo` we could have `const std::vector<const Detector *> detectors(size_t index)`.
+
+Note that although the index in the public interface is the same. They relate to two different indexing mechanisms. For example, on `SpectraInfo::isMonitor(size_t index)` the `index` is the spectrum index, which goes from 0 to the maximum number of spectra. On the other hand `DetectorInfo::isMonitor(size_t index)` the `index` is the detector index. This can be larger than the available number of spectra when mapping is not 1:1. 
+
+### Read access
+
+- Access functions take an index which is used for accessing an underlying `std::vector` of data.
+  - The data layout in `GeoemtryDataArray` is not defined, and could be changed in the future.
+  - To facilitate the free data layout, the interface should be `SpectrumInfo::twoTheta(size_t index)` and *not* along the lines of the current implementation with `Workspace::getDteector(size_t index)::twoTheta()`.
```
OwenArnold: I very much agree. The Facades should direct access to the `Detector` a little obscure, but not impossible. We certainly need to have clients prefer `SpectrumInfo::twoTheta(size_t index)`

Design/Instrument-2.0/design-notes.md
```diff
@@ -0,0 +1,89 @@
...
+  - Accessors should probably be declared inline for performance reasons.
+
+### Write access
+
+Write access requires modification of all three data structures, `InstrumentTree`, `DetectorInfo`, and `SpectrumInfo`.
+I (Simon) do not yet have an entirely clear picture of this.
+Notes:
+
+- Do we always go via `InstrumentTree` upon modification? For example, masking a spectrum or a detector could be handled at the `SpectrumInfo` or `DetectorInfo` level. These components could also have a copy-on-write mechanism. We would then still share the same `InstrumentTree` for workspaces that have different masking.
+- See the `Command` mechanism in the [prototype](https://github.com/DMSC-Instrument-Data/instrument-prototype). These could be filtered through `SpectrumInfo` and `DetectorInfo`.
```
OwenArnold: Yes, in fact I think it's essential that they are filtered through `SpectrumInfo` and `DetectorInfo`. This should be easy enough.

Design/Instrument-2.0/design-notes.md
```diff
@@ -0,0 +1,89 @@
...
+- When changing, e.g., the position of a bank, we will subsequently change the position of all pixels in that bank. We have to make sure that we do not trigger a full update of `SpectrumInfo` with each position change of a pixel. Instead, the derived position in `SpectrumInfo` should be computed only once after the update is complete.
```
OwenArnold: Almost certainly, the right way to focus on this requirement is to code it into the `Command`. The new `InstrumentTree` is only generated once the `Command` has been executed through the sub-tree. Likewise, the rebuild on the SpectrumInfo/DetectorInfo would only happen once that InstrumentTree had been rebuilt.

OwenArnold: @SimonHeybrock Another strategy that comes to mind would be to use version numbers to do precise updates. At present, when `InstrumentTree` is [modified](https://github.com/DMSC-Instrument-Data/instrument-prototype/blob/master/cow_instrument/Node.cpp#L57), all `Nodes` in it get an updated `version` number. This is simply used at the moment for testablity purposes. But we could easily change this so that only `Nodes` that where touched by the `Command` where incremented in version. Once the new `InstrumentTree` has been created, that would provide a way to only update the parts of the `DetectorInfo` that where affected by the change. On the other hand, this seems to be an unnecessary early optimisation.

OwenArnold:
@SimonHeybrock There where some additional things to discuss. After you respond, I'll update the document further.

Design/Instrument-2.0/design-notes.md
```diff
@@ -0,0 +1,89 @@
...
+
+The interface to `SpectrumInfo` and `DetectorInfo` would be almost identical.
+The most typical examples for public methods include:
+
+```cpp
+bool isMonitor(size_t index);
+bool isMasked(size_t index);
+double L2(size_t index);
+double twoTheta(size_t index);
+V3D position(size_t index);
```
quantumsteve: Does `position` have to be returned by value, or would it be better to return a const reference?

Design/Instrument-2.0/design-notes.md
SimonHeybrock: I think I agree, with a slight modification: Since we were discussing COW based on information in `DetectorInfo`, it might make sense to construct based on an `cow_ptr<InstrumentTree>` instead of a `unique_ptr`?

Design/Instrument-2.0/design-notes.md
SimonHeybrock: Not sure what is best design-wise. I would like to avoid having virtual-function resolution.

Maybe more importantly: We do not want clients to throw `DetectorInfo` and `SpectrumInfo` in the same bucket. According to my current understanding, they are two different things and *it should not be possible to threat them in the same way via a base class interface*.

Design/Instrument-2.0/design-notes.md
SimonHeybrock: I wrote this example rather carelessly, but having thought about it after your comment, I don't think the answer is so clear:

We argued somewhere that we do want to hide the underlying data layout. That could also mean that internally there is no `V3D` object, but only three vectors of `x`, `y`, and `z` coordinates, which then can be combined into a `V3D` on request. That data layout may have performance benefits (or not). Do you think it is feasible to keep the layout hidden from clients for now?

Design/Instrument-2.0/design-notes.md
OwenArnold: As an amendment to what I said above, as widely as possible the `InstrumentTree_uptr` would actually be a `InstrumentTree_const_uptr` to clarify the fact that the `InstrumentTree` if fully immutable. In the new design both the `InstrumentTree` and `Node` types are completely immutable. That makes the cow behaviour redundant at this level. The new design manages all modifications and copies internally, and these always result in the fabrication of a new `InstrumentTree`.  One thing we might wish to do at the `Workspace` level is to provide helper functionality to automatically swap the `uptr` post Modification. For example:

```cpp

MoveBankCommand command{...};
this->executeOnInstrument(command)

void Workspace::executeOnInstrument(const Command& command){
  // Ensure that the modified instrument is assigned against the workspace post command.execute
  auto newInstrument = m_instrument->modify(command);
  m_instrument.swap(newInstrument);
}

```

SimonHeybrock: But the same `InstrumentTree` could still be shared by several `DetectorInfo` instances, couldn't it?

Design/Instrument-2.0/design-notes.md
OwenArnold: Noted. Yes, we can avoid the the virtual function resolution. Maybe we'll keep these two types completely separate at the class level.

Design/Instrument-2.0/design-notes.md
OwenArnold: My example above is wrong. I need to fix it because we agreed that the `DetectorInfo` holds the `InstrumentTree` the Workspace does not.

OwenArnold: Given that the `Workspace` holds the `SpectrumInfo`, which holds the `DetectorInfo` which holds the `InstrumentTree`, we may wish to make the `SpectrumInfo` `COW` even though there is nothing to gain by making the `InstrumentTree` COW. `DetectorInfo` and `SpectrumInfo` hold `GeometryInfo`. This data may well be the same across multiple workspaces.

SimonHeybrock: Yes, `InstrumentTree` does not need to be COW (or rather it is, but via an internal mechanism and not simply by a `cow_ptr`). However, it can be the same accross multiple `DetectorInfo` objects, i.e., it should be held via a `shared_ptr`, not a `unique_ptr`.

OwenArnold: Yes I think this could be done via a shared_ptr.

Design/Instrument-2.0/design-notes.md
OwenArnold: I'm not aware of the use-case that would drive us to expose these arrays directly? This would seem to go against the encapsulation we gain in the `V3D`, with it's useful member functions.

SimonHeybrock: It would be merely for performance reasons. The data layout would much better suit the vector units of the CPU. For now, it is unclear if and where this will be necessary, what I am saying is just that I would like to keep our options open, unless it severely impacts other code (this is what we need to figure out).

For most clients the vectors should nevertheless be hidden, as you say.

Design/Instrument-2.0/design-notes.md
SimonHeybrock: @OwenArnold To avoid duplication, I was just discussing the 'CRTP' with Michael. He is facing a similar problem in some HKL redesign, and I am in the design of the new Histogram type. See also http://eli.thegreenplace.net/2011/05/17/the-curiously-recurring-template-pattern-in-c.

Design/Instrument-2.0/design-notes.md
OwenArnold: I can see that vectors may be a better choice, particularly where positions need to be looped over. I think the `V3D` is still important to our high-level client code. Some options might be:

* Expose the `V3D` on the `InstrumentTree` anyway, but also add  the ability to get x, y, z as vectors 
* Do not present a `V3D` in layers lower down than the `SpectrumInfo`

Design/Instrument-2.0/design-notes.md
OwenArnold: @SimonHeybrock Yes,, I am familiar with it. We could use it here as it would give no virtual function lookup, provide the code sharing we want, but not imply any runtime polymorphic relationship between the two `Info` types.

Design/Instrument-2.0/design-notes.md
quantumsteve: It might be worth looking into VTK's recent changes to allow either SOA or AOS memory layouts.
http://public.kitware.com/pipermail/vtk-developers/2016-March/033282.html

OwenArnold: @quantumsteve thanks for pointing [this](https://gitlab.kitware.com/vtk/vtk/blob/master/Documentation/Doxygen/ArrayDispatch-VTK-7-1.md) out. The SOA approach seems to make a lot of sense, and given that VTK have been able to take generic compile-time approaches to handling these is encouraging. However, I have several issues with taking the SOA approach when it comes to V3D.

* We have situations where we want to take a polymorphic approach to `Components`, of which `Detectors` are a subclass. Currently `Detectors` are "discovered" and registered at runtime in to the `InstrumentTree` as a AOS (see [here](https://github.com/DMSC-Instrument-Data/instrument-prototype/blob/master/cow_instrument/InstrumentTree.h#L37)). To do the equivalent using the SOA approach, and maintain the ability to return an individual `Detector` such as is done [here](https://github.com/DMSC-Instrument-Data/instrument-prototype/blob/master/cow_instrument/InstrumentTree.h#L26) we'd need some way to slice through our SOA and create the `Detectors` on the fly. 
* Certain geometric operations are still the responsibility of `Components` such as positioning and rotating. Adding two `V3D`s together seems like a sensible operation and provides a nice encapsulation. The same would be true of `Quat` for the equivalent rotation matrix. Furthermore the new `InstrumentTree` relies on knowing and walking sub-trees for write operations. How does one handle this in a SOA configuration. Currently each detector has an index, so it is possible to walk the tree and index into the SOA.

What are peoples thoughts?

SimonHeybrock: I suppose that many of our operations on `V3D` are limited by memory bandwidth. In that case SOA does not help much (unless you need, e.g., *only* the y-coordinate). If there is some really have Geometry calculation that is not bandwidth-bound the conversion to and from SOA could be done on-the-fly before and after that calculation (with some overhead).
All in all, the advantages of SOA in practice (in Mantid) are right now probably few, so I agree that we should not inflict other headaches upon us, i.e., I would agree with keeping the current AOS approach.

OwenArnold:
@SimonHeybrock @quantumsteve Thanks for your comments. I'll edit the document according to what has been discussed. I believe there is enough information here now for me to prototype the high-level stack ontop of the `InstrumentTree` now, so I will do so in the new repository.

PR #18, opened 2016-04-15
Design for HKLs

MichaelWedel:
This is the current draft of a design for type safe HKLs. Please add comments as you see fit, I'd appreciate some feedback.

### DIFF
--- a/Design/Crystallography/HKL.md
+++ b/Design/Crystallography/HKL.md
@@ -105,7 +105,7 @@ NumericType indexSumOfDifference(const IsHKL<NumericType, Derived> &lhs, const I
 Suggested implementation
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-There's a [PR]() against the main repository where this design is implemented. It revealed one unfortunate name clash, `Mantid::Geometry::HKL` already exists as a sub-class of `MDFrame`. My suggestion for this is to use the opportunity and start migrating the crystallography related code into a new module called Crystallography, which then provides a new namespace `Mantid::Crystallography`. Initially I suggest that this new module sits between `Kernel` and `Geometry`. Once the existing crystallography-related code has been migrated, the dependency of the Geometry module on the Crystallography module can be removed.
+There's a [PR](https://github.com/mantidproject/mantid/pull/15914) against the main repository where this design is implemented. It revealed one unfortunate name clash, `Mantid::Geometry::HKL` already exists as a sub-class of `MDFrame`. My suggestion for this is to use the opportunity and start migrating the crystallography related code into a new module called Crystallography, which then provides a new namespace `Mantid::Crystallography`. Initially I suggest that this new module sits between `Kernel` and `Geometry`. Once the existing crystallography-related code has been migrated, the dependency of the Geometry module on the Crystallography module can be removed.
 
 In the pull request there is a performance test as well that compares the performance for some common operations that are performed on HKLs between the different cases `int` and `double`, but also with `V3D`.
 
--- a/Design/Crystallography/HKL.md
+++ b/Design/Crystallography/HKL.md
@@ -1,0 +1,134 @@
+Making HKL different
+====================
+
+Motivation
+----------
+
+Currently the class `Kernel::V3D` is used to represent a large range of points and vectors in different spaces. Detector coordinates, unit cell vectors, Miller indices and many more are all represented by the same very low-level class `V3D`. While this is certainly very flexible in terms of the number of classes and/or interfaces developers need to keep track of, it comes with a number of disadvantages. The most important is probably that it's possible to perform nonsensical operations such as adding detector coordinates to a unit cell vector - since different types do not exist, they can not be enforced by the compiler, which means the burden of preventing errors falls to the developer. A similar case is where real and reciprocal space vectors have to be transformed by using different matrices. Without dedicated types the developer needs to make sure the correct matrix is used and in addition runtime checks have to be performed.
+
+While it is a big task to make each and every use of `V3D` (and connected to that also `Kernel::Matrix`) type-safe, implementing a dedicated type for Miller indices HKL is a rather confined task. Potential replacements for `V3D` in this domain can exist in parallel at first and code can be ported step-by-step, simply reverting to the current "unsafe" implementation where necessary.
+
+Requirements
+------------
+
+There should be a datatype different from `V3D` but with similar capabilities that is used to represent Miller index triplets.
+
+While in theory, Miller indices are defined to be composed of three integer numbers, some flexibility is required for cases in early steps of data processing where exact integers may not be known or for commensurate superstructures where fractional indices can occur.
+
+Though similar, those three cases cover different uses and should be covered by different types:
+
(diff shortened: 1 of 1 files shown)


Design/Crystallography/HKL.md
```diff
@@ -0,0 +1,121 @@
+# Making HKL different
+
+## Motivation
+
+Currently the class `Kernel::V3D` is used to represent a large range of points and vectors in different spaces. Detector coordinates, unit cell vectors, Miller indices and many more are all represented by the same very low-level class `V3D`. While this is certainly very flexible in terms of the number of classes and/or interfaces developers need to keep track of, it comes with a number of disadvantages. The most important is probably that it's possible to perform nonsensical operations such as adding detector coordinates to a unit cell vector - since different types do not exist, they can not be enforced by the compiler, which means the burden of preventing errors falls to the developer. A similar case is where real and reciprocal space vectors have to be transformed by using different matrices. Without dedicated types the developer needs to make sure the correct matrix is used and in addition runtime checks have to be performed.
+
```
OwenArnold: I agree with the motivation for this.

Design/Crystallography/HKL.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Remaining issues
+
+Some open questions remain:
+
+  - Python export - not sure how well this transports through to Python
+  - Moving to new Crystallography-module
+  - Implement a "real world" example from the Geometry- or Crystal-module.
+
```
OwenArnold: * What about runtime-considerations. For example, `MDEvent` is a templated type, but various algorithms need to get dimension information at runtime. 
* How do we persist these `IsHKL` derived CRTP instances? How do we know what type they are upon reloading? 
* Do we extend typesafety to `H`, `K`, and `L`?

MichaelWedel: My thoughts in reverse order:

- I'm not sure how much gain there would be in creating own types for H, K, L separately - they are already dimensionless numbers, so basically integers, floats...
- This is a very good point, I'll look into combining these unambiguously with `std` streams. I suppose the question is whether it's necessary to persist single instances of these. At least if I think about saving/loading HKLs from files it's generally a list of those - that would be much easier to implement and more efficient.
- I'm not sure what you mean by runtime considerations - right now I don't see anything that `V3D` offers at runtime that these types don't.

Design/Crystallography/HKL.md
```diff
@@ -0,0 +1,121 @@
...
+There should be a datatype different from `V3D` but with similar capabilities that is used to represent Miller index triplets.
+
+While in theory, Miller indices are defined to be composed of three integer numbers, some flexibility is required for cases in early steps of data processing where exact integers may not be known or for commensurate superstructures where fractional indices can occur.
+
+Though similar, those three cases cover different uses and should be covered by different types:
+
+  1. `HKL` (integer, "proper Miller indices")
+  2. `ProHKL` (double, "early stages of reduction")
+  3. `FractionalHKL` (double, use for magnetic or other super-structures)
```
quantumsteve: Is `FractionalHKL` the right word to describe an incommensurate magnetic structure? Do we want to rename this type to something more general (`RealHKL`?, `DecimalHKL`?) and add another type for commensurate structures?

Would we gain anything by using a [rational number library](http://www.boost.org/doc/libs/1_60_0/libs/rational/rational.html) for `FractionalHKL`?

Design/Crystallography/HKL.md
```diff
@@ -0,0 +1,121 @@
...
+  
+  bool operator==(const Derived &rhs) {
+}
+};
+
+class ProHKL : IsHKL<double, ProHKL> {
+};
+
+class FractionalHKL : IsHKL<double, FractionalHKL> {
+};
```
quantumsteve: Why should `ProHKL` and `FractionalHKL` be double when `coord_t` is a float?

Design/Crystallography/HKL.md
MichaelWedel: You're right, we could use truly rational numbers for commensurate cases, `boost::rational` is already used for `SymmetryOperation` for example. One would need to keep one eye on performance though, because `boost::rational` does things like calculating the GCD and so on in some (or all of them? not entirely sure) of the arithmetic operations.

Design/Crystallography/HKL.md
MichaelWedel: The same question might go for `int` vs. `short`. I mean do we really expect unit cells so large that we need to be able to index to 2 billion?

There are two reasons I selected `double` for these types. First, `V3D` is clearly `double`, so I didn't think of using any other type. Second, I thought `Kernel::Matrix` wouldn't work with anything besides `int` and `double` (because it does not work with `short`) - a `grep` revealed that I was wrong about this.

Im indifferent to this - what are the gains of switching to `float`?

Design/Crystallography/HKL.md
OwenArnold: To be more specific about my first point. By "runtime considerations" I was really eluding to the runtime construction of these various types. Presumably we don't want to have switch statements everywhere, so how will construction be handled?

MichaelWedel: One way would be to use `HKLGenerator` for constructing these types. I added that a few months ago, currently it lets you generate a vector of `V3D`s. It could be made a template (with default `V3D` so current code doesn't break) so you'd write something along these lines:

```
HKLGenerator<IntegerHKL> g(unitCell, resolution);

std::vector<IntegerHKL> hkls;
std::copy(g.cbegin(), g.cend(), hkls.begin());
```

Did you have any other use cases in mind?

Design/Crystallography/HKL.md
SimonHeybrock: `float` is half the size, so you need half the memory bandwidth and half the cache size.

Design/Crystallography/HKL.md
```diff
@@ -0,0 +1,121 @@
...
+## Requirements
+
+There should be a datatype different from `V3D` but with similar capabilities that is used to represent Miller index triplets.
+
+While in theory, Miller indices are defined to be composed of three integer numbers, some flexibility is required for cases in early steps of data processing where exact integers may not be known or for commensurate superstructures where fractional indices can occur.
+
+Though similar, those three cases cover different uses and should be covered by different types:
+
+  1. `HKL` (integer, "proper Miller indices")
+  2. `ProHKL` (double, "early stages of reduction")
```
SimonHeybrock: I would prefer calling this `ProtoHKL`. `ProHKL` sounds a bit unclear?

Design/Crystallography/HKL.md
quantumsteve: Also, four floats fit into an SSE register. Are there places that would benefit substantially from something like [`AlignedVector3`](http://eigen.tuxfamily.org/dox/unsupported/group__AlignedVector3__Module.html)?

MichaelWedel: I suppose they could, but the underlying storage type is completely hidden away (except kind of for the iterators, but those could be provided in a different way I suppose), so this can (and should) be swapped out later. I would consider that a separate issue.

PR #22, opened 2016-07-15
Y E data storage modes

OwenArnold:
A point of discussion concerning Histogram Data.

### DIFF
--- a/Design/y_e_data_storage_modes.md
+++ b/Design/y_e_data_storage_modes.md
@@ -1,0 +1,133 @@
+# Y and E data storage modes
+
+## Uncertainties
+
+### Status
+
+- Typically, `E` stores the standard deviation.
+- For computations such as adding data, the variance is required, thus `E` is squared and converted back to standard deviation by taking the `sqrt` at a later points.
+  Often this is done in-place, i.e., `E` temporarily stores the variance.
+- In a few cases `E` is kept as a variance until just before the end of an algorithm for performance reasons.
+  For example, `DiffractionFocussing2` needs to summing many spectra.
+  To avoid repeated evaluation of the expensive `sqrt` function for the target spectrum, all spectra are added without converting back to standard deviation after every addition.
+
+Notable examples include `Kernel::VectorHelper::rebin()` and `Kernel::VectorHelper::rebinHistogram()`.
+
+### Discussion
+
+For `E` data, the modification of the storage mode is typically very local and a fix should be rather easy.
+
+- `Histogram` provides functions for obtaining variances as well as standard deviations, e.g., `countVariances()` and `countStandardDeviations()`.
+  If the internal `E` vector were converted to variances by external means (as it is done currently), these access methods would do the wrong thing.
+
+- `HistogramData` provides means for converting between standard deviations and variances, in particular types that do the conversion automatically.
+  These should be used by client code.
+
+- If summing spectra were done using `Histogram::operator+()`, propagating the uncertainties would involve a `sqrt` computation for every `+` operation.
+  In the example of `DiffractionFocussing2` this could yield a significant overhead, and it is not obvious how to solve this:
+
+  - The algorithm could do the summing itself, summing `Y` and variances separately.
+    This would cause no performance loss, and the code would be reasonably clean, e.g.,
+
+    ```cpp
+    for (size_t i=0; i<inputWS.getNumberHistograms(); ++i) {
+      counts += inputWS.counts(i);
+      variances += inputWS.countVariances(i);
+    }
+    ```
+
+  - We could provide optimized means for summing many spectra as part of the `HistogramData` module.
+    For example,
+
+    ```cpp
+    namespace HistogramData {
+    Histogram sumHistograms(const std::vector<Histogram> &histograms);
+    }
+    ```
+
+    This would be a clean solution without performance loss.
+    Unfortunately, there will be an implicit performance and ressource issue if the input is an `EventWorkspace`!
+    Obtaining a `Histogram` from the input workspace will create the `Y` and `E` data on the fly (this needs to be done in any solution).
+    However, to create the vector of histograms as input for `sumHistograms()` we need to **store all histograms at the same time**.
+    This will increase memory consumption and reduce cache locality of the computation.
+
+
+## Count and frequency data
+
+### Status
+
+- Data in `Workspace2D` be represent counts or frequencies.
+- Conversions between the two are done with the algorithms `ConvertToDistribution` and `ConvertFromDistribution`, or the helper function `WorkspaceHelpers::makeDistribution()`.
+- `MatrixWorkspace::isDistribution()` can be used for checking the state, and `MatrixWorkspace::setDistribution()` for setting the corresponding flag (note that this does not actually transform the data!).
+- There is a validator that verifies the state of a workspace, `RawCountValidator`, however is is used by relatively few algorithms.
+- A considerable number of algorithms use `MatrixWorkspace::isDistribution()` to adapt their behavior, in most cases this adaption is a simple conversion to the desired data storage type.
+- For a large number of algorithms there are no checks, and it is not documented whether or not the work only with counts, only with frequencies, or both.
+
+### Discussion
+
+If we permit storing either counts or frequencies
+
+1. `Histogram` can be in two states `YMode::Counts` and `YMode::Frequencies`.
+  `Histogram::counts()` and `Histogram::frequencies()` will always be able to return the correct thing.
+  However, `Histogram::y()` will reference data that has no fixed interpretation that client code can rely on.
+  To make this option work, we would thus need to refactor all client code to either use the `counts()`/`frequencies()` interface, or check the storage mode before using `y()`.
+
+2. `Histogram` can be in two states `YMode::Counts` and `YMode::Frequencies`.
+  `Histogram::counts()` and `Histogram::frequencies()` will always be able to return the correct thing.
+  However, `Histogram::y()` would reference data that has no fixed interpretation that client code can rely on.
+  We could thus remove this part of the interface and refactor all client code to use the `counts()`/`frequencies()` interface.
+
+3. `Histogram` has only a single state and always stores counts.
+  It will always be guaranteed that `Histogram::counts()` and `Histogram::frequencies()` return the correct thing, and client code can rely on `Histogram::y()` returning a reference to a vector of counts.
+  What about data that has no underlying counts, for example a function defined at certain points?
+  Maybe we need a new workspace type for this?
+
+
+The different interpretations of `Y` data effectively breaks the promises that the `Histogram` interface makes:
+
+- If we allow converting the internal `Y` data between counts and frequencies by clients, `Histogram::counts()` and `Histogram::frequencies()` will not return what their name implies
+
+- `RawCountValidator` can this be removed?
+
+- Saving data, e.g., `SEQUOIAreduction.py`
+
+- `FitMW` has flag to normalize, but does so only if input is histogram data. Ignores `isDistribution()`? But it seems to be inaccessible and is always off?
+
+- `MatrixWorkspace::isDistribution()` example `CreatePSDBleedMask` needs distribution and manually converts
+
+- binary operations
+
+
+- `Integration` always computes the integral of the counts, not the frequencies, even if the workspace stores distribution data.
+
+- Need to be able to load legacy files, in particular `LoadNexusProcessed`.
+
+- `MatrixWorkspace::YUnitLabel()` extra label for distribution?
+
+- `FractionalRebinning` and `RebinnedOutput` workspace?
+
+- how do we deal with point-distribution data? storing counts will be lossy (see maybe `IQTransform`).
+
+### Data visualization
+
+#### Plotting
+
+- It must be possible to plot either counts or frequencies
+
+`QwtWorkspaceSpectrumData` checks `MatrixWorkspace::isDistribution()`
+
+#### Table View
+
+If we do no longer store data as frequencies, it should still be possible to display the data in a histogram as frequencies.
+This could be done via a simple change in the table view, options are:
+
+1. Provide a way to toggle `Y` between counts and frequencies
+2. Replace `Y` by two columns, `Counts` and `Frequencies` (indicating that the latter is simply computed based on `Counts`, maybe via a tool tip).
+
+
+### Fitting
+
+- Do we ever want to fit counts?
+- If yes, we need to be able to select what `Fit` does.
+
+
--- a/Design/y_e_data_storage_modes.md
+++ b/Design/y_e_data_storage_modes.md
@@ -15,7 +15,7 @@
 
 For `E` data, the modification of the storage mode is typically very local and a fix should be rather easy.
 
-- `Histogram` provides functions for obtaining variances as well as standard deviations, e.g., `countVariances()` and `countStandardDeviations()`.
+- `Histogram` currently provides functions for obtaining variances as well as standard deviations, e.g., `countVariances()` and `countStandardDeviations()`.
   If the internal `E` vector were converted to variances by external means (as it is done currently), these access methods would do the wrong thing.
 
 - `HistogramData` provides means for converting between standard deviations and variances, in particular types that do the conversion automatically.
--- a/Design/y_e_data_storage_modes.md
+++ b/Design/y_e_data_storage_modes.md
@@ -9,9 +9,7 @@
   Often this is done in-place, i.e., `E` temporarily stores the variance.
 - In a few cases `E` is kept as a variance until just before the end of an algorithm for performance reasons.
   For example, `DiffractionFocussing2` needs to summing many spectra.
-  To avoid repeated evaluation of the expensive `sqrt` function for the target spectrum, all spectra are added without converting back to standard deviation after every addition.
-
-Notable examples include `Kernel::VectorHelper::rebin()` and `Kernel::VectorHelper::rebinHistogram()`.
+  To avoid repeated evaluation of the expensive `sqrt` function for the target spectrum, all spectra are added without converting back to standard deviation after every addition. `DiffractionFocusing2.cpp` calls `Kernel::VectorHelper::rebinHistogram()` where calculations between intput and output E `vector` are performed in terms of variance. This also happens in `Kernel::VectorHelper::rebin()` 
 
 ### Discussion
 
--- a/Design/y_e_data_storage_modes.md
+++ b/Design/y_e_data_storage_modes.md
@@ -67,21 +67,42 @@ For `E` data, the modification of the storage mode is typically very local and a
 
 If we permit storing either counts or frequencies
 
-1. `Histogram` can be in two states `YMode::Counts` and `YMode::Frequencies`.
-  `Histogram::counts()` and `Histogram::frequencies()` will always be able to return the correct thing.
-  However, `Histogram::y()` will reference data that has no fixed interpretation that client code can rely on.
-  To make this option work, we would thus need to refactor all client code to either use the `counts()`/`frequencies()` interface, or check the storage mode before using `y()`.
+1. **Current solution**:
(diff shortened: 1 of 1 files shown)


Design/y_e_data_storage_modes.md
```diff
@@ -0,0 +1,151 @@
...
+- In a few cases `E` is kept as a variance until just before the end of an algorithm for performance reasons.
+  For example, `DiffractionFocussing2` needs to summing many spectra.
+  To avoid repeated evaluation of the expensive `sqrt` function for the target spectrum, all spectra are added without converting back to standard deviation after every addition. `DiffractionFocusing2.cpp` calls `Kernel::VectorHelper::rebinHistogram()` where calculations between intput and output E `vector` are performed in terms of variance. This also happens in `Kernel::VectorHelper::rebin()` 
+
+### Discussion
+
+For `E` data, the modification of the storage mode is typically very local and a fix should be rather easy.
+
+- `Histogram` currently provides functions for obtaining variances as well as standard deviations, e.g., `countVariances()` and `countStandardDeviations()`.
+  If the internal `E` vector were converted to variances by external means (as it is done currently), these access methods would do the wrong thing.
```
OwenArnold: @SimonHeybrock  By this point. Do you mean that if the client has direct access to the E vector, they can essentially corrupt the storage vector, and therefore the runtime-conversion methods will not provide the right results?

SimonHeybrock: @OwenArnold Precisely! You would essentially scale with the inverse bin width twice if you do an external conversion and then call `frequencies()`.

Design/y_e_data_storage_modes.md
```diff
@@ -0,0 +1,151 @@
...
+    Histogram sumHistograms(const std::vector<Histogram> &histograms);
+    }
+    ```
+
+    This would be a clean solution without performance loss.
+    Unfortunately, there will be an implicit performance and ressource issue if the input is an `EventWorkspace`!
+    Obtaining a `Histogram` from the input workspace will create the `Y` and `E` data on the fly (this needs to be done in any solution).
+    However, to create the vector of histograms as input for `sumHistograms()` we need to **store all histograms at the same time**.
+    This will increase memory consumption and reduce cache locality of the computation.
+
```
OwenArnold: @SimonHeybrock. It sounds like you've answered this one. I don't see a way around the `EventWorkspace` issue. Are you proposing that the `Histogram::operator+()` is still provided even though it would be sub-optimal for the `DiffractionFocusing` scenario, and that the `sumHistograms` non-member would be provided alongside?

SimonHeybrock: @OwenArnold I think I favor providing only `operator+` and not `sumHistograms()`. In the special cases where performance matters (like `DiffractionFocussing`) it should not be much trouble to hand-write the addition (still simpler than it is now, since we will have `CountVariances::operator+=`).

SimonHeybrock: @OwenArnold On second thought, the `sqrt` thing might actually turn out to be a non-issue. `DiffractionFocussing` also needs to rebin each histogram and I think it is likely that the `sqrt` is negligible compared to that. I will test this once we do the change.

Design/y_e_data_storage_modes.md
OwenArnold: So it sounds like the solution should involve, not providing access to the raw stored `vector<double>` anywhere via the `Histogram` in order to make all of these operations safe. I think that's part of our rollout plan anyway. Eliminating the usage of these legacy api aspects is the first step towards deleting them. Am I missing anything in this?

Design/y_e_data_storage_modes.md
OwenArnold: @SimonHeybrock. Fine, but I think your example has highlighted something important. Understanding the storage mode for the `E` values (which we hide) is important to understand for clients using the `Histogram`.  For example, in your sum Histograms example, if the storage mode for the `E` value was variances instead of standard deviations, the user would not see the same performance problems. So my concern with operator+ is will work, but not necessarily well for certain problems.

Design/y_e_data_storage_modes.md
SimonHeybrock: Well, removing the legacy API will ensure that clients cannot break the *length* of the underlying vectors. But that does not help with this issue, which concerns the values of the vectors. The new interface still provides access via `mutableE()` and `setSharedE()`, and it will be quite hard to eliminate that.

Design/y_e_data_storage_modes.md
OwenArnold: To make my point clearer. It implicitly works, but this may not be a good thing.

SimonHeybrock: @OwenArnold You are right, I agree. But I don't know if we can do anything about that unless we completely remove direct assess to `E`.

Design/y_e_data_storage_modes.md
```diff
@@ -0,0 +1,151 @@
...
+### Status
+
+- Data in `Workspace2D` be represent counts or frequencies.
+- Conversions between the two are done with the algorithms `ConvertToDistribution` and `ConvertFromDistribution`, or the helper function `WorkspaceHelpers::makeDistribution()`.
+- `MatrixWorkspace::isDistribution()` can be used for checking the state, and `MatrixWorkspace::setDistribution()` for setting the corresponding flag (note that this does not actually transform the data!).
+- There is a validator that verifies the state of a workspace, `RawCountValidator`, however is is used by relatively few algorithms.
+- A considerable number of algorithms use `MatrixWorkspace::isDistribution()` to adapt their behavior, in most cases this adaption is a simple conversion to the desired data storage type.
+- For a large number of algorithms there are no checks, and it is not documented whether or not the work only with counts, only with frequencies, or both.
+
```
OwenArnold: An additional problem is that I think it  is conceptually incorrect for the whole workspace to be marked as either containing distributions or not. This is a per `Histogram` feature. I'm going to try to create a `python` simple api example of how I can corrupt this Algorithms scoped check...

SimonHeybrock: Well, that is not the only example, what is in my opinion much worse is `bool MatrixWorkspace::isCommonBins()` (don't look at it if you want to avoid some nightmares).

SimonHeybrock: @OwenArnold But yes, this is another reason for removing the `distribution` concept.

Design/y_e_data_storage_modes.md
OwenArnold: In that case full protection enforced at the APIL level will not be possible. 

Something that the standard committee did when introducing the more modern c++ usages, is to make the dangerous  features undesirable and deliberately ugly looking. Take for example a const cast: `const_cast<int&>(i)` or `reinterpret_cast` compared to the c-style casts. I'm wondering if the method for the mutable access ought to look more dangerous than `mutableE()`. Such as `unsafeE()`?

Design/y_e_data_storage_modes.md
OwenArnold: OK. I won't look at the implementation, but these seem like manifestations of the same problem. What I'm wondering if whether we can get away from the workspace informing us, and relying on those flags being set correctly. It would be a breaking change, but If users were explicitly forced to tell us whether the data was `frequency` or `counts` at the point of import (for example in `CreateWorkspace`, `Load` etc), per Historgram, then Mantid could be made to manage the consistency internally.

Design/y_e_data_storage_modes.md
SimonHeybrock: @OwenArnold Not convinced. `mutableE` is already a step forward from `dataE`.

Maybe we should just take things a step at a time. If we now eliminate all uses of conversions between variances in standard deviations in client code, we will see how things look then, and how they develop.
It is not clear that there is a big benefit from making direct access hard or removing it entirely, so we might be wasting effort.

Design/y_e_data_storage_modes.md
SimonHeybrock: @OwenArnold Yes, Ithink the points of import are easy to deal with. The flag can be removed if we define that histograms always store counts. My only concern is the lossy conversion to frequencies for point-distribution data.

We might consider adding an extra field to `Histogram` that stores a helper value that aids in making the conversion lossless:

- We always store counts in `Histogram`
- Client code needs point-data with frequencies, sue `Histogram(Points{1,2,3}, Frequencies{4,5,6}).
- To convert to `Counts` for the internal store, we need a bin-width, which is not uniquely defined since we have `Points`.
- Instead of lossy conversion (Mantid is currently taking the mid-points between points, but when you convert back to points you won't get the same):
  - convert to bin-edges using mid-points (just like now)
  - store points and counts, and the leftmost bin edge (the latter is not accessible on the public interface)
  - `Histogram::frequencies()` can now convert the counts back to frequencies on a *lossless* way, since we have the extra required value (in contrast to now)

SimonHeybrock: @OwenArnold What I described can guarantee that if you *start* with frequencies, then `frequencies()` always gives correct values, even if stored as counts internally. If you start with `Points` and `Counts`, `frequencies()` will still involve lossy conversion, but that is a mathematical fact that cannot be circumvented.

Design/y_e_data_storage_modes.md
```diff
@@ -0,0 +1,151 @@
...
+### Discussion
+
+If we permit storing either counts or frequencies
+
+1. **Current solution**:
+   - Data in `Histogram` can be in two states counts or frequencies, but the histogram does not know its state.
+   - `Histogram::counts()` and `Histogram::frequencies()` will thus not return the correct thing if the workspace stores distribution data.
+   - `Histogram::y()` references data that has no fixed interpretation that client code can rely on.
```
OwenArnold: I don't understand why we provide the unqualified `HistogramY` like this? This isn't part of the legacy API.

Design/y_e_data_storage_modes.md
```diff
@@ -0,0 +1,151 @@
...
+
+2. **Switchable Y-storage mode with direct Y access**
+   - `Histogram` can be in two states `YMode::Counts` and `YMode::Frequencies`.
+   - `Histogram::counts()` and `Histogram::frequencies()` will always be able to return the correct thing.
+   - However, `Histogram::y()` will reference data that has no fixed interpretation that client code can rely on.
+   - To make this option work, we would thus need to refactor all client code to either use the `counts()`/`frequencies()` interface, or check the storage mode before using `y()`.
```
OwenArnold: Forcing the client to use either `counts()` or `frequencies()` sounds like the right thing to do. Why do you thin that the safety would be **small-medium**? I agree, effort would be major.

Design/y_e_data_storage_modes.md
```diff
@@ -0,0 +1,151 @@
...
+
+3. **Switchable Y-storage mode without direct Y access**
+   - `Histogram` can be in two states `YMode::Counts` and `YMode::Frequencies`.
+   - `Histogram::counts()` and `Histogram::frequencies()` will always be able to return the correct thing.
+   - However, `Histogram::y()` would reference data that has no fixed interpretation that client code can rely on.
+   - We could thus remove this part of the interface and refactor all client code to use the `counts()`/`frequencies()` interface.
+   - Effort: **major**
+   - Benefit: **medium - high** (safe and versatile, we can store any data we may need)
+   - Safety: **high** (cannot misinterpret data, but it would be possible to run any algorithm on any type of data even if that would not make sense)
+
```
OwenArnold: 3. Is probably my favourite approach.

Design/y_e_data_storage_modes.md
SimonHeybrock: @OwenArnold It is direct access, but not the legacy API. It is slightly better, since it is type-safe and forbids length modifications.

Three reasons we need it:
- Some client code does not care if it works with `Counts` or `Frequencies`, so we cannot force it to use either of those.
- In-place modifications. Using the new interface this is not possible, since we return a COW object by value, so modification will trigger a copy, making in-place modifications of workspaces inefficient.
- The issues discussed in this document, i.e., we have places where data is converted between various storage modes, so for now client code cannot always rely on `counts()` and `frequencies()` to return what they need.

We will solve the last point and at least parts of the first point if we come to a good solution for what is being discussed here.

Design/y_e_data_storage_modes.md
```diff
@@ -0,0 +1,151 @@
...
+4. **Fixed Y-storage mode (counts)**
+   - `Histogram` has only a single state and always stores counts.
+   - It will always be guaranteed that `Histogram::counts()` and `Histogram::frequencies()` return the correct thing, and client code can rely on `Histogram::y()` returning a reference to a vector of counts.
+   - What about data that has no underlying counts, for example a function defined at certain points?
+     It would basically be converted into fake counts, based on a potentially lossy conversion of `Points` to `BinEdges`.
+     Maybe we need a new workspace type for this?
+   - Effort: **medium**
+   - Benefit: **medium** (safe but not versatile, we can only store counts and need workarounds for non-count data)
+   - Safety: **medium - high** (medium if we use `Histogram` to also store distribution data, converted to counts)
+
```
OwenArnold: If your storage mode is always `counts` you have the problem that you've described above with the E-data mode always being std dev. If you're data should be treated as frequencies you're penalty is a lot of on-the-fly conversions.

Design/y_e_data_storage_modes.md
SimonHeybrock: @OwenArnold See 3.), in 2.) we would still have direct access.

Design/y_e_data_storage_modes.md
```diff
@@ -0,0 +1,151 @@
...
+All of these options seem a bit awkward.
+This is also related to the fact that we can convert between `Points` and `BinEdges`.
+Do we need to fix that as well?
+Would it make sense to reduce this to a 3-state system, instead of having independent `XMode` and `YMode`, we would have `StorageMode::CountBins`, `StorageMode::FrequencyBins`, and `StorageMode::Points`, the latter implying that the data represents frequencies?
+
```
OwenArnold: Interesting proposition. But since `ConvertToPointData` algorithm does not produce frequencies, I think the single mode `StorageMode::Points` may not be sufficient for all existing use cases.

Design/y_e_data_storage_modes.md
SimonHeybrock: @OwenArnold I think you will have the same problem also if the storage mode is switchable. If you access frequency data via `counts()` you will also pay the conversion overhead. The only way to really get rid of the overhead would be to have two separate types and workspace types. But that is far beyond what we can do now, I think.

Design/y_e_data_storage_modes.md
```diff
@@ -0,0 +1,151 @@
...
+The different interpretations of `Y` data effectively breaks the promises that the `Histogram` interface makes:
+
+- If we allow converting the internal `Y` data between counts and frequencies by clients, `Histogram::counts()` and `Histogram::frequencies()` will not return what their name implies
+
+- `RawCountValidator` can this be removed?
```
OwenArnold: You can if you can safely convert frequencies->counts. The conversion logic is fine, but this goes back to the knowing what you have. This validator seems to just be checking the flag on the workspace `isDistribution`. As I mentioned previously, if we can properly import data into Mantid, I think we can eliminate the need for this workspace level flag, and also the validator.

On a slightly separate note. I think we've agreed that `isDistribution` and `isCommonBins` do not make sense at the `MatrixWorkspace` level. Now instead having a method `isAllDistributions` and `isAllCommonBins` might be useful as that's a cross-Histogram query?

SimonHeybrock: @OwenArnold I am not quite sure what you are saying. There error introduced by the `Frequencies` to `Counts` conversion is usually marginal unless you have quickly varying bin widths.

I don't think it makes sense to have some histograms in a workspace as a distribution while others are not. That will break 99% of our algorithms.

Design/y_e_data_storage_modes.md
OwenArnold: The second point you describe sounds like the only solid use-case we'd want to carry on going forward. 1 & 3, as you say are connected to this wider problem. 

You say: 'Some client code does not care if it works with Counts or Frequencies'. I can't think of an example where this is not important.

SimonHeybrock: @OwenArnold For example `ExtractSpectra`or other algorithms that simply chop off bits at the ends of the histogram. Also, I think, `ConvertUnits` does not care in many cases.

SimonHeybrock: @OwenArnold If we have switchable storage modes the direct access would also be necessary to preserve to the mode and preserve sharing:
```cpp
outWS.setCounts(inWS.counts());
```
If `inWS` stores frequencies sharing is broken (that may be desired or unintentional).

Design/y_e_data_storage_modes.md
OwenArnold: Agreed. 

As a side note, it would be nice to know aprori how the Historam was intending to be used. A separate workspace type would be one way of forcing the client to make a choice, (although they would only differ in the actual storage mode and would keep exactly the same API). Anyway, I think this is beyond what we can do with our available resources.

Design/y_e_data_storage_modes.md
OwenArnold: Indeed. Most algorithms would break and I can't think of a good scenario where you would desire the mix. What I currently dislike is that there are two sources of information that are supposed to be synchronized. The Histograms contain the definitive information about what we are storing. We have a separate flag at the workspace level that is disconnected from the Histograms.

Design/y_e_data_storage_modes.md
OwenArnold: `ExtractSpectra` or `ConvertUnits` may not care at the calculation level, but subsequent algorithms in a chain will need to know that what they are storing. So direct Y axis in the cases you have mentioned is OK, but the Histograms of the OutputWorkspaces will need to be created/set in a way where the information is preserved.

SimonHeybrock: @OwenArnold In the case of `X`, where we have two storage modes, the `Histogram` constructor takes care of this. That is, since basically all algorithms create their output workspaces via `WorkspaceFactory::create(parent)` this is taken care of automatically.

Design/y_e_data_storage_modes.md
OwenArnold: Agreed. You need this missing information to perform the lossless conversion for the `Points`. I'd be interested in seeing this fleshed out in terms of the proposed code changes.

Design/y_e_data_storage_modes.md
OwenArnold: @SimonHeybrock. In that case, for those algorithms there is no problem.

What i've suggested regarding the E, access and now the Y axis here is that the direct access is potentially unsafe and will lead to bugs that will be very hard to track down, because you won't get failure at the point of misuse. I would prefer that clients are forced to decide what they are getting either by means of making the direct access methods unavailable. My secondary approach (as I mentioned above) is to make such access non-standard looking.

OwenArnold: I appreciate your points regarding the practicalities of removing such methods in one go. I'm not at all proposing that we take a giant leap like that.

SimonHeybrock: @OwenArnold I think we can realistically do that only by moving a lot of functionality into the `HistogramData` scope (or a new module for `Histogram` operations). Basically most algorithms perform a per-histogram operation. These operations could be extracted from the algorithm code and moved into a histogram module. The algorithms themselves would then only perform operations with histograms, not with the X, Y, and E data.

This is again a massive effort, but it can easily done step by step over a longer period. That way we will eventually get to the point where the direct access interface is used little.

OwenArnold: @SimonHeybrock. I agree with this ethos.

1. Most algorithms should operate at the Histogram level.
1. Code relating to the internal manipulation of histograms should be migrated into the `HistogramData` scope
1. Access to aspects of the Histograms should be increasingly qualified. `X`, `Y`, `E` are internal.
1. We should aim to make these improvements as part of a longer programme of work.

PR #23, opened 2016-08-02
Release planning for Mantid version 4.0

NickDraper:
A design document to capture the discussions around Mantid version 4.0

### DIFF
--- a/Design/Mantid4ReleasePlanning.md
+++ b/Design/Mantid4ReleasePlanning.md
@@ -26,14 +26,17 @@ List of breaking changes
 ------------------------
 
 1. Change from qwt5 graphs to ? [3 people 1-3m]
+ 42. Expose `Histogram` object to python to simplify the task [1 person 1m]
 1. Sanitise Algorithm and property names (including merging) [Plan - 1 person 2w, Impl 3 people 1-3m]
  1. Remove unused algorithms, and old unused algorithm versions
 1. Removing InOut workspace usage [1 person 2w]
 2. Python API Changes
  1. ?Moving the Python API from the SimpleAPI (the impacts of this need further consideration) [????????]
  2. Remove the old Python interface to workspaces, only provide the new one (removing, e.g.,  readY()  and  dataY() ).
- 3. Remove Indexing breaking operations -  e.g ISpectrum::setSpectrumNo() 
- 1. Move to Python 3 on all platforms, removal of Python 2.7 support. [1 person 2m]
+ 3. Remove Indexing breaking operations -  e.g ISpectrum::setSpectrumNo()
+1. Move to Python 3 on all platforms
+ 42. Finish adding forward compatibility [2 people 1m]
+ 42. Removal of Python 2.7 support. [1 person 2m]
 
 List of other changes
 ---------------------
--- a/Design/Mantid4ReleasePlanning.md
+++ b/Design/Mantid4ReleasePlanning.md
@@ -25,18 +25,15 @@ We have not raised the concept of Mantid 4.0 with our users, so we have not made
 List of breaking changes 
 ------------------------
 
-1. Change from qwt5 graphs to ? [3 people 1-3m]
- 42. Expose `Histogram` object to python to simplify the task [1 person 1m]
-1. Sanitise Algorithm and property names (including merging) [Plan - 1 person 2w, Impl 3 people 1-3m]
+1. Sanitise Algorithm and property names (including merging & generating naming convention) [2 person 8 months]
  1. Remove unused algorithms, and old unused algorithm versions
-1. Removing InOut workspace usage [1 person 2w]
+ 1. Consistent naming of indexing properties: spectrum numbers, workspace indices, detector IDs  [1 person 6 months]
+ 1. Removing InOut workspace usage [1 person 2w]
+
 2. Python API Changes
+ 1. Python exports to HistogramData [1 person 1 month]
+ 2. Remove the old Python interface to workspaces, only provide the new one (removing, e.g.,  readY()  and  dataY() ). [1 person 1 month]
  1. ?Moving the Python API from the SimpleAPI (the impacts of this need further consideration) [????????]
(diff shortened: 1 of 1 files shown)


Design/Mantid4ReleasePlanning.md
```diff
@@ -0,0 +1,65 @@
...
+User Expectations of Mantid 4.0
+-------------------------------
+
+We have not raised the concept of Mantid 4.0 with our users, so we have not made any commitment about what it may contain.  However we should be aware that while it is  not defined in Semantic versioning (http://semver.org/), the users will expect some significant improvements in order to justify the version change, and to act as a carrot to entice them to upgrade and overcome the discomfort of the breaking changes we are introducing.  Some of the breaking changes themselves will bring significant advantages, but the benefits of others are less visible.  It is worth considering other significant improvements that can be aligned with Mantid 4.0 to encourage enthusiastic adoption of these changes.
+
+List of breaking changes 
+------------------------
+
+1. Change from qwt5 graphs to ? [3 people 1-3m]
+1. Move to Python 3 on all platforms [1 person 2m]
```
quantumsteve: Would you be more specific about what this change includes. Our current strategy is to maintain python 2.7 and python 3.x compatibility. Will there be a transition period where we ship separate packages for python 2 & 3? Can we realistically drop ALL support for python 2 with alongside the other substantial changes outlined here?

quantumsteve: Building RHEL 7 RPMs for all Mantid dependencies using python 3 is a significant undertaking. @stuartcampbell might have a better idea of the time required.

Design/Mantid4ReleasePlanning.md
```diff
@@ -0,0 +1,65 @@
...
+1. Sanitise Algorithm and property names (including merging) [Plan - 1 person 2w, Impl 3 people 1-3m]
+1. Removing InOut workspace usage [1 person 2w]
+1. ?Moving the Python API from the SimpleAPI (the impacts of this need further consideration) [????????]
```
SimonHeybrock: - Remove old algorithm versions?

SimonHeybrock: - Remove the old Python interface to workspaces, only provide the new one (removing, e.g., `readY()` and `dataY()`).

SimonHeybrock: - It is likely that changes to indexing (https://github.com/mantidproject/documents/pull/24) will require removing things like `ISpectrum::setSpectrumNo()`, i.e., there is a series of breaking changes. This is a prerequisite for Instrument 2.0.

Design/Mantid4ReleasePlanning.md
```diff
@@ -0,0 +1,65 @@
...
+
+List of other changes
+---------------------
+
+These changes are not bound to be included in this release, but being significant changes it would be good to include them together if
+the timing and resource allows.
+
+1. Instrument 2.0 [2p 4m]
```
SimonHeybrock: - I think 2p 4m might be lower bound for roll out of the indexing changes I mention above, probably more, depending on the number of complications that arise.
- Currently I think it is likely that as a follow up and before we can introduce instrument 2.0 we might also need to cleanup workspace creation and initialization. Maybe 2p 2-4m.
- Instrument 2.0 itself can easily take 2p 6m for a basic implementation. This does *not* include advanced functionality, and does *not* include removing the old interface, i.e, refactoring existing algorithms comes on top of that (apart from simple cases).

OwenArnold: I think 2p 6m for instrument 2.0 is a reasonable guess for `Instrument 2.0`. I would however much prefer that this estimate is kept private to the development team until the prototype is finished. I think I'll be able to give a much better estimate when we have approximations for all the features that we need built into the prototype.

NickDraper:
Don't panic too much about the estimates at the moment for the PMB this afternoon I'll share a version with them with all of the estimates removed.

NickDraper:
The PMB approved of the concept.  We need to flesh out what we plan to do, evaluate what should or should not be done for each area, the effort to implement, what approaches we can take to mitigate the change and the costs of these.

Design/Mantid4ReleasePlanning.md
```diff
@@ -0,0 +1,77 @@
...
+
+User Expectations of Mantid 4.0
+-------------------------------
+
+We have not raised the concept of Mantid 4.0 with our users, so we have not made any commitment about what it may contain.  However we should be aware that while it is  not defined in Semantic versioning (http://semver.org/), the users will expect some significant improvements in order to justify the version change, and to act as a carrot to entice them to upgrade and overcome the discomfort of the breaking changes we are introducing.  Some of the breaking changes themselves will bring significant advantages, but the benefits of others are less visible.  It is worth considering other significant improvements that can be aligned with Mantid 4.0 to encourage enthusiastic adoption of these changes.
+
+List of breaking changes 
+------------------------
+
+1. Sanitise Algorithm and property names (including merging & generating naming convention) [2 person 8 months]
```
quantumsteve: Exposing C++ getters/setters to python as [class properties](http://www.boost.org/doc/libs/1_63_0/libs/python/doc/html/tutorial/tutorial/exposing.html#tutorial.exposing.class_properties) would make the interface more pythonic. PR #18350 shows some of the opportunities & challenges.

OwenArnold:
@NickDraper Why isn't this being merged. Don't we want to keep the outputs/decisions?

NickDraper:
@OwenArnold Primarily this was killed by the user meeting of Jan 2017, or at least drastically changed when there was strong objection to almost all of the breaking changes.  There will not and has not been implemented in any particularly reasonable state. So it would seem strange to merge it as such.

OwenArnold:
@NickDraper As you say, many of these changes were rejected in Jan 2017. Is there a separate record of that? I'm worried that this is the only document that would contain a record of what was originally proposed.

PR #24, opened 2016-08-03
Design document: Indexing module

SimonHeybrock:
Design document about an indexing module, which is a prerequisite for various other changes. Please comment!

There was a previous PR for this, but the document has been completely revised and significantly extended. Some old comments can be found there, most of them have been included in an updated version of the document: https://github.com/mantidproject/documents/pull/13.


Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,269 @@
...
+There are a few more complex cases.
+For example, in `LoadEventNexus` a vector of events labeled by detector ID needs to be inserted into a workspace.
+This requires *filtering* a vector with data based on a spectrum number of detector ID.
+That is, we would not return a set of indices but a vector that is a subset of the input vector, i.e., the input vector with all data corresponding to other MPI ranks removed.
+
+#### Implementation and effects on existing code
+
+- Implement the `Translator` and store it in a workspace.
+  Most likely it makes sense to share this between workspaces, i.e., use `Kernel::cow_ptr`, since indices change rarely after initial creation.
+  As either part of the `Translator`, or as a separate entity within `MatrixWorkspace`, we will have a `Partitioning` object.
```
MichaelWedel: This is tricky, from what I understand this depends mainly on how, where and when the "global index" is kept. Could it be a subclass of `Translator`, an `MPITranslator` or something like that?

Moreover, at what point is the partitioning supplied to the `Translator` and where does it come from?

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,269 @@
...
+  This object is responsible for determining how data of spectra is split among MPI ranks.
+  There can be several implementations of this `Partitioning` to fit the needs of various work flows.
```
MichaelWedel: Can this change for a single workspace between algorithms? Maybe (also see above) `Partioning` could have something like `Translator Partitioning::transformTranslator(const Translator &t);`, which takes a `Translator` object and creates a new one based on that, so `Partitioning` basically becomes a "source" of `Translator` in a specific MPI-partitioning scenario. It could leave the spectrum numbers untouched and just attach a global workspace index mapping.

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,269 @@
...
+  `Translator` uses `Partitioning` to set up its internal maps that define which spectrum numbers are present on a specific MPI rank.
```
MichaelWedel: Based on my thoughts above, I think it could be like this:

- `MatrixWorkspace` has `Translator` and `Partitioning` members.
- There are different `Partionings`, including a "default" one for no MPI/one rank.
- `MatrixWorkspace::setTranslator(....)` could, before storing, adapt the translator to the desired partitioning scheme using the partitioner, along the lines of `m_translator = m_partitioner.transformTranslator(userSuppliedTranslator);`.

Disadvantage: `MatrixWorkspace::getTranslator` does not necessarily return what you put into `MatrixWorkspace::setTranslator`.

Design/spectrum_number_and_workspace_index_abstraction.md
SimonHeybrock: @MichaelWedel I am not sure yet. My initial implementation can be found here: https://github.com/mantidproject/mantid/pull/16836. However this is currently just an isolated solution, i.e., I have not yet attempted to unify this with creation of workspaces. As you can see, the `Partitioning` is passed to the constructor of the `Translator` (in that PR there are two separate translator, `SpectrumNumberTranslator` and `DetectorIDTranslator`, it is not clear yet if the would be unified into a single object or not).

The main problem, as I just discussed with @OwenArnold, is that we need to globally define what the current `Partitioning` is. All algorithms must use the same `Partitioning`, otherwise everything will break in an MPI run. And we would like to avoid singletons...

Design/spectrum_number_and_workspace_index_abstraction.md
SimonHeybrock: @MichaelWedel Sounds mostly good 👍 . Note that we do not (necessarily) need to distinguish between `Partitioning` with and without MPI. You just set the partition count to 1.

Design/spectrum_number_and_workspace_index_abstraction.md
OwenArnold: @MichaelWedel. For the purposes of this design. How the internals of the `Partitioning` are  set is not important. All that we need to know is that `Partitioning` can be decided aprori and injected. 

We are not sure what the best type of partitioing will be yet, it could be technique dependent, but it's the critical piece in achieving good load balancing, and it's something we want to keep flexible.

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,269 @@
+# Indexing Abstraction Layer
```
FedeMPouzols: Nice diagram. This is the only place where "spectrum index" appears. Maybe that needs to be explained somewhere?
It seems it is not a Mantid concept, but as that wording is used in NeXus files it is probably good to have its meaning clearly defined here. 
I'm not totally sure if "spectrum index" is equivalent to "workspace index". Does it always increase by one? Is that stated/guaranteed anywhere?

I think that increasing by one has been the common practice at ISIS, but recently we have seen files where they break that assumption for some prototype detectors.

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,269 @@
...
+    This is simple to do, an example can be found as part of [this pull request](https://github.com/mantidproject/mantid/pull/15465)
+    I believe this will cover 95% of cases and ensure that there will be no performance hit.
+  - In the other cases, when an actual list of spectra is given, I suppose that that list is commonly small (since it is often manual user input), so `map` access is negligible.
+  - The remaining performance critical case is loading event data. Each event has an associated detector ID, and this needs to be translated to a workspace index.
+    This is currently done by creating a `vector` (instead of a `map`) for look-up.
+    We certainly have to support a similar mechanism.
+    This is a case where we will need to write some performance tests, to prevent regression.
+    These tests should be added well in advance so we can gather enough statistics prior to any changes and monitor the introduction of the changes.
+
+To deal with MPI in a way that is hidden from both the user and most developers, `Translator` transforms a user input into, e.g., a `SpectrumIndexSet`.
```
martyngigg: Can we avoid the use of the phrase "spectrum index" and use only "spectrum number" or "workspace index"? I think it makes it easier to distinguish the two.

Design/spectrum_number_and_workspace_index_abstraction.md
SimonHeybrock: @FedeMPouzols I consider the term "workspace index" confusing (after all it is not an index of a workspace in, say, a list of workspaces), so this was just an attempt to sneak in alternative terminology. But maybe we just have to stick with the old one?

My intention was to use "spectrum index" instead of "workspace index", i.e., it is just a memory offset and has nothing to do with working in NeXus files.

OwenArnold: If we haven't already. Maybe it is worth stating some of the guarantees about what `index` means. integer set starting from 0 etc. @SimonHeybrock I prefer the new terminology. Yes "workspace index" can be a confusing one, particularly when working with groups.

Design/spectrum_number_and_workspace_index_abstraction.md
SimonHeybrock: @martyngigg Would you also suggest to pick a different name for "detector index", to distinguish it easier from "detector ID"?

Design/spectrum_number_and_workspace_index_abstraction.md
martyngigg: Ideally, I think workspace/spectra index needs to be banished from the user interface (both graphical/non-graphical) so that it's purely a developer thing and then I think we have more scope for renaming the concept.

SimonHeybrock: We can rename the concept and keep the naming in the UI the same (if it gets not banished). As I wrote somewhere, what the UI sees is a global index, which in any case for an MPI run is not that same as the current "workspace index".

FedeMPouzols: I very much agree that not having workspace/spectra index in any UI is the ideal.
A point to consider though is that there can be workspaces with rows that are not spectra, for example the outputs from Fit or other algorithms which are just matrices with rows of data.

SimonHeybrock: @FedeMPouzols Well, but even if the rows are not spectra we still have `MatrixWorkspace::getSpectrum()` that returns an `ISpectrum` (with data that is not a spectrum). If we really want to solve this we need new workspace types.

Design/spectrum_number_and_workspace_index_abstraction.md
```diff
@@ -0,0 +1,269 @@
...
+# rank 0: indices contains 1
+# rank 1: indices contains 0,1
+# rank 2: indices is empty
+```
+
+Algorithms then use:
+
+```python
+for index in indices:
+  ws.mutableY(index) += offset
```
quantumsteve: Would it be possible to (and make sense to) make more objects iterable? Perhaps something like

```python
for value in ws.mutableY()
    value += offset
```

SimonHeybrock: `mutableY(index)` returns a `HistogramY` at that workspace index. `HistogramY` is iterable, so you can do:

```cpp
for (const auto & y : ws.mutableY(index))
  sumY += y;
```
We have also considered avoiding the explicit loop over `indices` by providing something like `MatrixWorkspace::apply(const SpectrumIndexSet &indices, T operator)` (or maybe as part of `Algorithm`, so we can do cancelling and progress reporting).

SimonHeybrock:
The basics have been settled, this can be updated in another PR if there a significant changes, so I am merging this for now (mainly because I want to link to it from another document).

PR #27, opened 2016-09-01
Indexing rollout notes

SimonHeybrock:
In progress, this PR is meant for discussion.

### DIFF
--- a/Design/indexing_rollout_notes.md
+++ b/Design/indexing_rollout_notes.md
@@ -51,9 +51,9 @@ Furthermore there are comparisons between the respective spectrum and detector i
   ```
 
 - Should we set default detector IDs?
-  This is was workspaces are doing currently, but it does not really make sense.
+  This is what workspaces are doing currently, but it does not really make sense.
   Probably we should find a better way if setting sensible detector IDs once an instrument has been set.
-  This is something that potentially needs to be decided together with other decisions regarding Instrument-2.0.
+  This is something that potentially needs to be planned together with other decisions regarding Instrument-2.0.
 
 
 ## NearestNeighbours
@@ -116,7 +116,8 @@ if (numDetectors > 1) {
 ## Duplicate spectrum numbers
 
 Currently quite a few algorithms or tests use or create workspaces with non-unique spectrum numbers.
-Ultimately we want to forbid this, but it is not yet clear if we should attempt to clean this up right away, or wait until `IndexInfo` has developed into a more stable state.
+Ultimately we want to forbid this, but we should avoid changing too many things at once.
+This should be cleaned up once `IndexInfo` has developed into a more stable state.
 
 
 ## Standards
@@ -166,7 +167,7 @@ It would probably be easier and cleaner to simply remove special handling of spe
 
 #### MDWorkspace
 
-`ConvertToDiffractionMDWorkspace::convertEventList` takes an `EventList` and its its events to an `MDWorkspace`.
+`ConvertToDiffractionMDWorkspace::convertEventList` takes an `EventList` and adds its events to an `MDWorkspace`.
 This procedure requires the position of the `EventList`.
 Currently this is determined by the detector IDs stored in the `EventList`.
 A different mechanism is required here, probably with a change in the function signature.
@@ -207,3 +208,6 @@ We seem to have two options:
   The disadvantage of this are potential performance issues.
 
 2. Simply replace the exports by dummy versions that print errors containing information on how to fix scripts.
+
+Realistically, we will have to go with option 1, then add exports for the new Python interface and deprecate the old interface.
+In a later release we can then either directly remove the old interface, or use option 2 as yet another intermediate step, before actually removing the old exports.
--- a/Design/indexing_rollout_notes.md
+++ b/Design/indexing_rollout_notes.md
@@ -14,9 +14,9 @@ Note that this does not necessarily imply that the nomenclature on user-interfac
 ## Statistics
 
 Most relevant methods that need to be removed are part of `ISpectrum`.
-The following table gives and estimate of the number of occurences of the respective methods.
+The following table gives and estimate of the number of occurrences of the respective methods.
 
-| Method        | Occurences | Files  |
+| Method        | Occurrences | Files  |
 | ------------- |:-------------:|:-----:|
 | `ISpectrum::getSpectrumNo()` | 271 | 99 |
 | `ISpectrum::setSpectrumNo()` | 153 | 104 |
@@ -76,6 +76,12 @@ Furthermore there are comparisons between the respective spectrum and detector i
   }
   ```
 
+## File- and UI output
+
+Similar to logging, there are also several algorithms that write, e.g., detector IDs to a file.
+Maybe a similar or even the same helper function as suggested above in the logging section could be used.
+
+
 ## Obtaining detector information
 
 Frequently, detector IDs are used to obtain detector pointers or other detector-related information from the instrument.
@@ -90,6 +96,7 @@ class SpectrumInfo {
   bool isMonitor(const size_t index) const;
   double l2(const size_t index) const;
   Detector_sptr uniqueDetector(const size_t index) const; // Throws if no or more than 1 detector
+  bool hasUniqueDetector(const size_t index) const; // Should this be part of IndexInfo instead?
 };
 ```
 
@@ -123,6 +130,28 @@ Should this be cleaned up as part of this refactoring effort?
 Examples:
 
 - `AppendSpectra`
+
+
+## MatrixWorkspace
+
+There is a considerable number of methods in `MatrixWorkspace` that could potentially be replaced or removed during this refactoring effort.
+The detailed strategy should depend on how and where these methods are used.
+
+```cpp
+class MatrixWorkspace {
+public:
+  void updateSpectraUsing(const SpectrumDetectorMapping &map);
+  void rebuildSpectraMapping(const bool includeMonitors);
+  spec2index_map getSpectrumToWorkspaceIndexMap() const;
+  detid2index_map getDetectorIDToWorkspaceIndexMap(bool throwIfMultipleDets) const;
+  std::vector<size_t> getDetectorIDToWorkspaceIndexVector(detid_t &offset, bool throwIfMultipleDets) const;
+  std::vector<size_t> getSpectrumToWorkspaceIndexVector(specnum_t &offset) const;
+  std::vector<size_t> getIndicesFromSpectra(const std::vector<specnum_t> &spectraList) const;
+  size_t getIndexFromSpectrumNumber(const specnum_t specNo) const;
+  std::vector<size_t> getIndicesFromDetectorIDs(const std::vector<detid_t> &detIdList) const;
+  std::vector<specnum_t> getSpectraFromDetectorIDs(const std::vector<detid_t> &detIdList) const;
+};
+```
 
 
 ## Notes about individual algorithms
--- a/Design/indexing_rollout_notes.md
+++ b/Design/indexing_rollout_notes.md
@@ -1,6 +1,6 @@
 # Indexing rollout
 
-This is not a design document.
+This is not a design document, but an addendum to the [Indexing design document](https://github.com/mantidproject/documents/blob/master/Design/spectrum_number_and_workspace_index_abstraction.md).
 This document contains notes and is meant for planning the rollout of indexing changes, in particular removing spectrum number and detector IDs from `ISpectrum`.
 
 ## Remarks
--- a/Design/indexing_rollout_notes.md
+++ b/Design/indexing_rollout_notes.md
@@ -50,7 +50,10 @@ Furthermore there are comparisons between the respective spectrum and detector i
   }
   ```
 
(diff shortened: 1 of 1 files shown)


Design/indexing_rollout_notes.md
```diff
@@ -0,0 +1,209 @@
...
+  That is, we should *not* just refactor the manipulation of spectrum numbers and detector IDs via `IndexInfo`.
+  Instead, we should provide functionality for grouping and extraction of spectra to create a new `IndexInfo`, without the need for explicit handling of spectrum numbers and detector IDs.
+
+  As an example, consider
+
+  ```cpp
+  namespace Indexing {
+  IndexInfo extract(const IndexInfo &, const SpectrumIndexSet &);
+  IndexInfo group(const IndexInfo &, const std::vector<SpectrumIndexSet> &);
+  }
```
OwenArnold: I like the idea of having non-member functions to help with this kind of thing. It removes the need to know how to operate on the `IndexInfo` and allows you to do all the allocation associated with the updated `IndexInfo` in one hit.

Design/indexing_rollout_notes.md
```diff
@@ -0,0 +1,209 @@
...
+  ```
+
+- Should we set default detector IDs?
```
OwenArnold: I think this is a general construction-order question. I have no idea what "sensible" detector_ids is. I think this knowledge needs to come from the client somehow, not something that we auto-generate.

Design/indexing_rollout_notes.md
```diff
@@ -0,0 +1,209 @@
...
+  This is was workspaces are doing currently, but it does not really make sense.
+  Probably we should find a better way if setting sensible detector IDs once an instrument has been set.
+  This is something that potentially needs to be decided together with other decisions regarding Instrument-2.0.
+
+
+## NearestNeighbours
+
+- Uses `specnum_t` as map key.
+- Should simply use spectrum index?
```
OwenArnold: I'm not sure that we can use spectrum index for exactly the reasons you describe. Global spectrum index might be OK.

Design/indexing_rollout_notes.md
```diff
@@ -0,0 +1,209 @@
...
+## Logging and exception messages
+
+- Frequently spectrum numbers are included in log messages or exceptions.
+
+- Probably the easiest (overall) solution is to add means of getting this information automatically, e.g., with a helper function
+
+  ```cpp
+  namespace Logging {
+  std::string spectrumNumber(const MatrixWorkspace &workspace, size_t index);
+  }
```
OwenArnold: Seems reasonable

Design/indexing_rollout_notes.md
```diff
@@ -0,0 +1,209 @@
...
+## File- and UI output
+
+Similar to logging, there are also several algorithms that write, e.g., detector IDs to a file.
+Maybe a similar or even the same helper function as suggested above in the logging section could be used.
+
+
+## Obtaining detector information
+
+Frequently, detector IDs are used to obtain detector pointers or other detector-related information from the instrument.
+The IDs are not actually required here.
```
OwenArnold: True in the case that the client is not specifying the detector id. Instrument 2.0 essentially does not use this information since most operations are applied in bulk to all detectors, therefore you only need to know how to enumerate over them. Passing a predicate to aid a search might be a useful option.

Design/indexing_rollout_notes.md
```diff
@@ -0,0 +1,209 @@
...
+Refactoring would be significantly easier (and would avoid the need to revisit and incrementally refactor) if such information was available via, e.g., `MatrixWorkspace::SpectrumInfo`.
+
+Commonly used functionality includes:
+
+```cpp
+class SpectrumInfo {
+  bool isMasked(const size_t index) const;
+  bool isMonitor(const size_t index) const;
+  double l2(const size_t index) const;
+  Detector_sptr uniqueDetector(const size_t index) const; // Throws if no or more than 1 detector
```
OwenArnold: Are you passing back a shared_ptr just to keep things compatible with what we already have?

Design/indexing_rollout_notes.md
```diff
@@ -0,0 +1,209 @@
...
+  det = inst->getDetector(*workspace->getSpectrum(wi).getDetectorIDs().begin());
+} else {
+  det = workspace->getDetector(wi);
+}
+```
+
+## Duplicate spectrum numbers
+
+Currently quite a few algorithms or tests use or create workspaces with non-unique spectrum numbers.
+Ultimately we want to forbid this, but it is not yet clear if we should attempt to clean this up right away, or wait until `IndexInfo` has developed into a more stable state.
```
OwenArnold: I suggest the latter approach. Not trying to change too many things at once. But I agree spectrum numbers should be unique.

Design/indexing_rollout_notes.md
```diff
@@ -0,0 +1,209 @@
...
+
+
+## Standards
+
+#### Spectrum numbers starting at 0
+
+Currently quite a few algorithms that create workspaces where spectrum numbers do not start at 1 but at 0.
```
OwenArnold: I'm not sure why this matters? We want to shift over to using spectrum indexes. Very few operations should use spectrum numbers, they should be a user thing.

Design/indexing_rollout_notes.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Notes about individual algorithms
+
+The following is a more or less random selection/collection about specific algorithms or classes and how the could or should change.
+
+#### SumSpectra
+
+Uses smallest spectrum number in sum as spectrum number for output.
+This does not seem to make much sense, maybe it is just an arbitrary requirement by an individual piece of client code.
+It would probably be easier and cleaner to simply remove special handling of spectrum numbers from `SumSpectra` and adapt the client code in question.
```
OwenArnold: This by itself sounds like a large task.

Design/indexing_rollout_notes.md
```diff
@@ -0,0 +1,209 @@
...
+
+#### MDWorkspace
+
+`ConvertToDiffractionMDWorkspace::convertEventList` takes an `EventList` and its its events to an `MDWorkspace`.
```
OwenArnold: repeat of "its"

Design/indexing_rollout_notes.md
```diff
@@ -0,0 +1,209 @@
...
+This procedure requires the position of the `EventList`.
+Currently this is determined by the detector IDs stored in the `EventList`.
+A different mechanism is required here, probably with a change in the function signature.
+
+#### SpectrumDetectorMapping
+
+This class is used in a few algorithms, e.g., `SofQWNormalisedPolygon`, to set/update the detector IDs in a workspace via `MatrixWorkspace::updateSpectraUsing`.
+Probably this functionality can simply be replaced by basic `IndexInfo` functionality.
+Ultimately, it might be possible to remove the class `SpectrumDetectorMapping`.
```
OwenArnold: Agreed. The `IndexInfo`needs to be the unique, consistent and unambiguous source of this information.

Design/indexing_rollout_notes.md
```diff
@@ -0,0 +1,209 @@
...
+
+1. Check if the code in question can be removed. This is the case, e.g., when setting spectrum numbers that either match the default or when the actual values do not matter (some tests).
+
+
+## Python exports
+
+The methods from the `ISpectrum` interface are exported to Python.
+They seem to be used in relatively few scripts, but we should make sure that scripts outside our repository keep working, or fail in a way that suggests a solution.
+We seem to have two options:
+
```
OwenArnold: I think we are going to have to support 1 below.

Design/indexing_rollout_notes.md
SimonHeybrock: I agree. We will have to see how it works out in practice. Basically what we want is a mechanism that (automatically?) sets sensible detector IDs if the workspace has an instrument (or when a new instrument ist set)?

Design/indexing_rollout_notes.md
SimonHeybrock: Well, it depends. If the spectrum index is only used as a map index to obtain all neighbors of *local indices* (the neighbors themselves may be non-local) it would be fine.

Design/indexing_rollout_notes.md
SimonHeybrock: Yes, I think there are 3 reasons:

- As long as `SpectrumInfo` is just a wrapper around Instrument-1.0 detectors do not exist permanently, but a parametrized detectors is allocated on demand. This can be buffered in `SpectrumInfo`. If we would return a reference there is some danger of invalidating it by requesting information for another index. That is, returning a `shared_ptr` makes this a bit safer. Later it could (and should) be changed to a reference.
- There might still be other client code that requires a `shared_ptr`, which we also need to clean up first.
- Copy and paste ;-)

Design/indexing_rollout_notes.md
SimonHeybrock: Well, there are algorithms that create workspaces without user input such as files. In particular, the example I gave (`AppendSpectra`) merged spectra from workspaces, but for some reason the spectrum numbers start at zero. That is, spectrum numbers actually are used directly in this algorithm, since it has to set them to a non-default value. If we would use defaults, we could probably eliminate most of such code.

Design/indexing_rollout_notes.md
SimonHeybrock: Actually it seems pretty simple. It makes the changes in `SumSpectra` itself trivial (just remove 10-15 lines of code instead of a more complex refactor that needs to directly interact with `IndexInfo`). Apart from `SumSpectraTest` there is only one other unit test that fails (some `Engg*` thing), so that algorithm might be the only one that needs fixing.

Design/indexing_rollout_notes.md
SimonHeybrock: Ok, I'll add a comment about multiple rollout stages.

Design/indexing_rollout_notes.md
OwenArnold: OK. As long as you keep in mind that we should never modify spectrum numbers away from what the client provides either via the load-data, or some setting as part of an algorithm.

Design/indexing_rollout_notes.md
OwenArnold: Ok, but there's also the non-automated, even by-eye testing. As in there will be different visible results if one say opens up the Matrix View in MantidPlot, or on down-stream user scripts. I have no real way of quantifying this.

Design/indexing_rollout_notes.md
SimonHeybrock: @OwenArnold In what I have implemented so far in https://github.com/mantidproject/mantid/pull/17394 I am actually not providing access to detectors. I think it would be used in some places, but I am not sure that this belongs into `SpectrumInfo`. I think the key question for this is:
*Does Instrument-2.0 provide/need detector groups?*
Currently detector groups are used for things like defining which detectors belong to a spectrum, and for passing it around as a single object, for reading parameters (which may get averaged for all detectors in the group). Is there anything in this list that Instrument-2.0 cannot solve better in another way?
- `SpectrumInfo` covers many of the cases. For example, we can get L2 for a spectrum, which is defined as the average of all associated detectors. This does not require groups.
- `ParameterMap` access? This is still unclear.
- Others?

If we do not have detector groups `SpectrumInfo` cannot and should not be able to return detector pointers or references. I might be able to return containers, with zero or more entries. One can argue that a detector group is just a container, but actually the composite pattern used in Instrument-1.0 is maybe not ideal here. For example, `IDetector::getID()` returns the detector ID of the first detector in the group. This seems wrong. Maybe forcing this into a common schema is just not the right thing here?

Design/indexing_rollout_notes.md
SimonHeybrock: @OwenArnold After having tried when implementing the fake `SpectrumInfo`: Changing this will be a really major effort. For example:

- Currently detector IDs are default generated.
- In many tests no instrument is set, or only a dummy instrument that does not have the corresponding detector IDs.
- Many algorithms (corresponding to those tests) attempt to get parameters (such as `isMasked`) for spectra, but skip if there is no detector.

I we now enforce correct detector IDs, the algorithm behavior will change, and thus the tests results. I have seen cases where this affects hard coded values, so it will be really tedious to fix this.

What I am saying is: Of course we still need to do this, but I am starting to think that it might be better to try and stage it more. That is, when first adding `IndexInfo`, we still allow bad IDs. Only in a later step we fix this.
I still need to figure out whether or not this approach will lead to other problems though.

OwenArnold: Agreed. On a project of this size, with this magnitude of changes, a staged approach is very sensible.

Design/indexing_rollout_notes.md
OwenArnold: I think that `SpectrumInfo`/`Instrument 2.0` should not have Detector Groups. For the reasons you say. I think the detector groups were simply a mix of the spectrum and detector concepts, hence why they are created on the fly. I don't think that `SpectrumInfo` should give you any direct access to detector things like ids.

SimonHeybrock:
No major changes to this for now, so I'll merge this. Can be updated later with no trouble since this is not a design document.

PR #31, opened 2016-10-11
SpectrumInfo rollout notes

SimonHeybrock:
@OwenArnold As discussed on Slack, here is a first draft. Let me know what you think.


OwenArnold:
Looks good to me. We could maybe think about translating (using this to advise) a Github project.

SimonHeybrock:
Good idea! I'll think a bit about how we can split this into as many small issues as feasible.

SimonHeybrock:
I'll merge this so we have a file to link to from issues. Can be updated later.

PR #36, opened 2016-12-01
Scanning instrument design document

ianbush:
This PR contains a design document for the implementation of scanning instruments. Also contains a requirements document based on needs at the ILL.


Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,102 @@
...
+
+The `timeIndex` argument is optional, so algorithms that do not care about time indexing will simply return the first/only time index. `DetectorInfo` should also contain a vector of start and end times for each time index. This will mean any information stored as a time series sample log can also be obtained for the correct step, for example the monitor counts for normalisation.
+
+`DetectorInfo` will ultimately store the information for positions and rotation, so will need to do this in a way that can be accessed by both the detector index and time index. Masking could also be stored in a time indexed fashion.
+
+#### `SpectrumInfo`
+
+Currently [`SpectrumInfo`](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/SpectrumInfo.h) has a `getDetector` method for finding the detector for the workspace. This uses the `getSpectrum(index)` method on `MatrixWorkspace` to return the detector IDs for the spectrum.
```
SimonHeybrock: Note: I think this is mainly a workaround that I did not intend to keep in the final implementation.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,102 @@
...
+
+A new method should be added, for example `getTimeIndexedDetectorIDs(index)` to return a set of `std::pair<detid_t, size_t>`.
```
SimonHeybrock: So this is for the case of grouping combined with scanning?

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,102 @@
...
+
+This can be used to access methods on `DetectorInfo`, for example `m_detectorInfo.twoTheta(detIndex, timeIndex)`.
+
+#### `MatrixWorkspace` & `ISpectrum`/`Spectrum`
+
+As metioned under **`SpectrumInfo`** `MatrixWorkspace` will require a `getTimeIndexedDetectorIDs(index)` method.
+
+This implies the member variable `std::set<detid_t> detectorIDs` in `ISpectrum` will need to become `std::set<std::pair<detid_t, size_t>> timeIndexedDetectorIDs`.
```
SimonHeybrock: I think you are probably right that this might be needed as a temporary workaround. Where possible, we should however try to eliminate the need to directly work with detector IDs. Once we have a fully functional `SpectrumInfo` and `DetectorInfo` there should be very few cases where algorithms need to detector IDs.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,102 @@
...
+
+The corresponding methods in `ISpectrum` would also need to be replicated for the time indexed case. For example `addDetectorID(const std::pair<detid_t, size_t> timeIndexedDetID)`.
```
SimonHeybrock: I agree. Again note that we ultimately want and need to remove detector IDs from `ISpectrum`. See also https://github.com/mantidproject/documents/blob/master/Design/spectrum_number_and_workspace_index_abstraction.md.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+The following methods will be given require a time index access, in addition to the detector index.
+
+```cpp
+DetectorInfo(boost::shared_ptr<const Geometry::Instrument> instrument,
+           Geometry::ParameterMap *pmap = nullptr);
+
+bool isMonitor(const size_t index) const;
+bool isMasked(const size_t index, const size_t timeIndex=0) const;
```
SimonHeybrock: I don't think masking should be time dependent. If only a specific position is to be masked that would be done by masking the spectrum.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,102 @@
...
+double l2(const size_t index, const size_t timeIndex=0) const;
```
SimonHeybrock: I wonder if having a default time index isn't dangerous. I had in mind having a routine without time index that throws if there is more than one time position of the detector. So there would be two variants:

```cpp
double l2(const size_t index) const; // used by algorithms that do not support scanning instruments, throws for scanning instruments
double l2(const size_t index, const size_t timeIndex=0) const;
```
Same for the other methods.

Design/Instrument-2.0/ScanningInstruments_ILL.md
```diff
@@ -0,0 +1,122 @@
...
+* Multiple workspaces can be merged, with different sets of time indexes, and positions and rotations stored correctly for each detector
+* Detector positions for a given time index can still be modified after loading
+* Detector movement at the ILL is limited to the case of rotation around sample position - but full positional changes should be allowed, for example in case the sample is off-centre
+* Flexible step scans are supported, times at a given position can differ, as can angular step size changes between steps
+* Each spectrum ultimately links to a detector or detector group and a time index for scanning instruments
+* A start and end time, as the full time and date, can be obtained for each time index
+* Extra metadata can be accessed for a given position in the step scan - for example so normalisation to monitors can be performed later
+* Metadata values that change in time can be shared appropriately between positions in the step scan (for example in D2B files each file contains a single value for monitor counts, but 25 different detector positions, so the monitor counts for each position might need to be divided by 25)
+* Overlaps in detector positions are possible
```
SimonHeybrock: Good point. I think there is no requirement on the instrument here? There might be certain algorithms that make trouble?

Design/Instrument-2.0/ScanningInstruments_ILL.md
```diff
@@ -0,0 +1,122 @@
...
+* The instrument view shows every step position for each detector, and deals with overlapping detectors
+* Saving/loading workspaces should deal with all of the information related to step scans
+
+### Desirable
+* Multiple workspaces can be merged with the same detector positions for certain time indexes - the data for spectra is then summed
+* Masking of detectors at different time indexes
```
SimonHeybrock: As mentioned above, I would handle this by two-level masking: Masking at the spectrum level and masking at the detector level.

Design/Instrument-2.0/ScanningInstruments_ILL.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Performance
+
+Instrument 2.0 is already providing a new framework for performance. The only requirement for the ILL to use scanning instruments is not to introduce any performance regressions. Otherwise the performance only needs to be the equivalent to that of having a larger instrument.
+
+D2B represents the most difficult case for performance at the ILL, with 100 steps x 64 detectors x 256 pixels = 1,638,400 spectra.
```
SimonHeybrock: Do you see any scenario where we might get above the int32 limit? If so we should make sure to go with int64 indices right from the start.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,102 @@
+## Scanning Instrument Support for the ILL
+
+### Background
+
+#### Instrument 2.0
+
+The full discussion on Instrument 2.0 is beyond the scope of this document, more details are contained in this [requirements document](requirements-v2.md). A short summary on [scanning instrument requirements is given](requirements-v2.md#moving-instruments-requirements) within this document.
+
+For the purposes of the discussion here scanning instruments (or moving instruments) describe instruments that have detectors that can be in different positions during a single run. This forms a step scan. This is used to cover data collection where there are gaps between detectors. The aim of the work is to be able to deal with such detectors that move during a run, and be able to accumulate the data into a single workspace with the detectors at different positions throughout the run. This does not include two other use cases from the Instrument 2.0 requirements document; continuous scans and triple axis spectrometers.
```
OwenArnold: Yes. Definition is right. For `Instrument 2.0` we would probably also want to consider scenarios where other components (non-detectors) could be scanned. I discussed this with @SimonHeybrock and our conclusion is that this should not be the focus of the immediate work and would very much be tied up in the full `Instrument 2.0` rollout itself.

Also. @MichaelWedel and I happen to think that `Instrument-ng` `ng` = next generation, might be a better name. Especially since I started versioning stages of the prototype.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,102 @@
...
+
+#### Overview of Proposed Implementation
+
+Two classes related to this have already been implemented in Mantid, [`SpectrumInfo`](https://github.com/mantidproject/mantid/pull/17394) and [`DetectorInfo`](https://github.com/mantidproject/mantid/pull/17875). These define the new way of working with commonly used information about a spectrum, namely isMasked, isMonitor, l2, twoTheta, signedTwoTheta, position, rotation and eFixed (coming soon). Eventually `DetectorInfo` will be responsible for the ownership of the parameters for the position and rotation, so for example moving a detector would have the new position stored in the `DetectorInfo` class, instead of the parameter map as is done currently. Values such as `l2` and `twoTheta` would use the new position when called and cache the return value, until another position change invalidates it.
```
OwenArnold: `implemented` = `partially implemented`.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,102 @@
...
+
+Support for the moving instruments will be implemented via the `DetectorInfo` class. The `SpectrumInfo` class needs no direct knowledge of the time indexing. The methods on the `DetectorInfo` class for accessing detector information will be accessed by giving a detector index and an optional time index. It will also hold information on the start and end times for each time index. Any further information, for example to normalise to monitor counts, can be held in a time series sample log.
```
OwenArnold: **Usage** of the `SpectrumInfo` class needs no direct knowledge of the time indexing.

Design/Instrument-2.0/ScanningInstruments_Design.md
ianbush: Yes, I agree that it makes more sense to mask at the spectrum level here.

Design/Instrument-2.0/ScanningInstruments_ILL.md
```diff
@@ -0,0 +1,122 @@
...
+* The instrument view shows every step position for each detector, and deals with overlapping detectors
+* Saving/loading workspaces should deal with all of the information related to step scans
+
+### Desirable
+* Multiple workspaces can be merged with the same detector positions for certain time indexes - the data for spectra is then summed
+* Masking of detectors at different time indexes
+
+### Performance
+
+Instrument 2.0 is already providing a new framework for performance. The only requirement for the ILL to use scanning instruments is not to introduce any performance regressions. Otherwise the performance only needs to be the equivalent to that of having a larger instrument.
```
OwenArnold: I think ```Instrument 2.0 is already providing a new framework for performance``` is a bit ambiguous. Also lets be very clear that anything "Instrument 2.0" like (see comment above) is a prototype. Not ready for production.

OwenArnold: But good to mention that key non-functional about performance.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,102 @@
...
+
+bool isMonitor(const size_t index) const;
+bool isMasked(const size_t index, const size_t timeIndex=0) const;
+double l2(const size_t index, const size_t timeIndex=0) const;
+double twoTheta(const size_t index, const size_t timeIndex=0) const;
+double signedTwoTheta(const size_t index, const size_t timeIndex=0) const;
+Kernel::V3D position(const size_t index, const size_t timeIndex=0) const;
+Kernel::Quat rotation(const size_t index, const size_t timeIndex=0) const;
+
+void setPosition(const size_t index, const Kernel::V3D &position, const size_t timeIndex=0);
```
OwenArnold: So is this the plan for construction. The `DetectorInfo` is made initially around a non-scanning instrument and additional detector scan positions are added via these setters? That could be quite a slow way of constructing a "scanning instrument"

Design/Instrument-2.0/ScanningInstruments_Design.md
ianbush: Yes, this makes sense to have the two variants. It would be dangerous if you were ignoring a whole bunch of detector positions.

Design/Instrument-2.0/ScanningInstruments_ILL.md
ianbush: Yes, you are right, I do not think there should be any requirements on the instrument for this. Other than the instrument view I am not aware of any problems doing this, but I am sure we will find some along the way.

Design/Instrument-2.0/ScanningInstruments_ILL.md
ianbush: For the ILL requirements that I know of I can not see that being required. How about at the ESS? If anything comes close we should probably go for int64 now.

Design/Instrument-2.0/ScanningInstruments_Design.md
ianbush: Yes, I am not sure if there is ever a reason to group in time, but as [`GroupDetectors`](http://docs.mantidproject.org/nightly/algorithms/GroupDetectors-v2.html) can work on spectra I guess that should be allowed too.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,115 @@
...
+
+Currently [`SpectrumInfo`](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/SpectrumInfo.h) has a private `getDetector` method for finding the detector for the workspace. This uses the `getSpectrum(index)` method on `MatrixWorkspace` to return the detector IDs for the spectrum.
+
+A new method should be added, for example `getTimeIndexedDetectorIDs(index)` to return a set of `std::pair<detid_t, size_t>`.
+
+This can be used to access methods on `DetectorInfo`, for example `m_detectorInfo.twoTheta(detIndex, timeIndex)`.
+
+#### `MatrixWorkspace` & `ISpectrum`/`Spectrum`
+
+As metioned under **`SpectrumInfo`** `MatrixWorkspace` will require a `getTimeIndexedDetectorIDs(index)` method.
```
OwenArnold: Might be nice to give the full signatures with named variables marking whether we are talking about detector index or time indexes. I think the current reviewers understand this well enough, but the document should have a wider audience.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,115 @@
...
+
+This implies the member variable `std::set<detid_t> detectorIDs` in `ISpectrum` will need to become `std::set<std::pair<detid_t, size_t>> timeIndexedDetectorIDs`.
+
+The corresponding methods in `ISpectrum` would also need to be replicated for the time indexed case. For example `addDetectorID(const std::pair<detid_t, size_t> timeIndexedDetID)`.
+
+The original methods such as `this->detectorIDs.insert(detID)` should remain and add detectors at time index 0. This allows existing algorithms, e.g. loaders, that do not need time indexing to work as before. Similarly the `getDetectorIDs(index)` method would still return detector IDs without the time indexing information.
+
+**Note:** working directly with the detector IDs is something that should be eliminated with the use of `SpectrumInfo` and `DetectorInfo`. The use of a `getTimeIndexedDetectorIDs` or `getDetectorIDs` should not be needed in the final Instrument 2.0. Similarly the method `getSpectrum(index)` on `MatrixWorkspace` should eventually be eliminated.
```
OwenArnold: Good point. Thanks for clarifying this for our readers.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Implementation
+
+For more details on the rollout plan see the document [Scanning Instruments Plan](ScanningInstruments_Plan.md).
+
+For this design to be implemented the use of `SpectrumInfo` and `DetectorInfo` will need to be rolled out in Mantid. The rollout for [`SpectrumInfo`](https://github.com/mantidproject/mantid/issues/17743) has been started.
```
OwenArnold: I think we should probably say a little more about the rollout to indicate that it's not an afternoons work. Both classes need to be rolled out across the entire framework. We need to intercept the majority of direct access to the `Instrument` via these caching layers. Even Algorithm authors should not routinely be dealing with directly with the `Instrument`.

OwenArnold:
@ianbush. I think this is very good. There are two things that I think are missing.

1. How the time dependent information is going to be stored in the `DetectorInfo`. I know that this is a little low-level, but I think it would be something that a reader might want to know. As a corollary to that, how (if at all) do we prevent clients from accessing static positions from the nested `Instrument`.
1. How any information in these caching layers might be persisted. Once you've built a scanning workspace. You're going to want to store it.
1. How the scanning `DetectorInfo` will be constructed. I think this is an important point. The semantics of this need to be right, and there are performance considerations as I mentioned before.

Design/Instrument-2.0/ScanningInstruments_Design.md
ianbush: Yes, I think this could make sense. So when creating the `SpectrumInfo` object you would need to pass a vector of std::set<std::pair<detid_t, size_t>> objects, for step scans?

ianbush:
@OwenArnold @SimonHeybrock I have pushed some changes to this that should address 3. and the first part of 1.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -18,6 +18,8 @@ Two classes related to this have already been partially implemented in Mantid, [
 
 Support for the moving instruments will be implemented via the `DetectorInfo` class. The usage `SpectrumInfo` class needs no direct knowledge of the time indexing. The methods on the `DetectorInfo` class for accessing detector information will be accessed by giving a detector index and an optional time index. It will also hold information on the start and end times for each time index. Any further information, for example to normalise to monitor counts, can be held in a time series sample log.
 
+Currently initialisation of workspaces is usually done in an incremental way. This will have performance impacts for the loader, but can also leave the workspace in an inconsistent state. For the scanning instrument design the workspace should be initialised all at once, via a `DetectorInfo` factory, which will ensure the loader creates a workspace in a consistent state.
```
OwenArnold: @ianbush Yes I agree. Could you be a bit more specific about what you mean about this for other readers (the first sentence).

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -28,33 +30,52 @@ The current plan is to access to add methods that take a detector index and a ti
...
 
 #### `SpectrumInfo`
 
 Currently [`SpectrumInfo`](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/SpectrumInfo.h) has a private `getDetector` method for finding the detector for the workspace. This uses the `getSpectrum(index)` method on `MatrixWorkspace` to return the detector IDs for the spectrum.
 
-A new method should be added, for example `getTimeIndexedDetectorIDs(index)` to return a set of `std::pair<detid_t, size_t>`.
+A new method should be added, for example `std::set<std::pair<detid_t, size_t>> getTimeIndexedDetectorIDs(index)`. The time indices for each detector can be stored in `SpectrumInfo` as a `std::set<std::pair<detid_t, size_t>>` within a vector of the same size as the number of workspace spectra. This will avoid the need to change methods on `MatrixWorkspace` or `ISpectrum/Spectrum` to return the set of detector IDs and time indexes for the spectrum. This is something that will help to move away from the incremental construction of `DetectorInfo`.
```
OwenArnold: Very good. I think this is a good improvement.

ianbush:
@OwenArnold @SimonHeybrock I have just pushed some more changes. These cover saving and loading step scans, instrument access and `SpectrumInfo` construction.

These are my initial thoughts on each topic, but they might need some discussion.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -18,7 +18,20 @@ Two classes related to this have already been partially implemented in Mantid, [
...
+  for (size_t j = 0; j < m_numberOfPixelsPerTube; ++j) {
+    int *data_p = &data(static_cast<int>(i), static_cast<int>(j), 0);
+    m_localWorkspace->setHistogram(spec, m_localWorkspace->binEdges(0),
+        Counts(data_p, data_p + m_numberOfChannels));
+    spec++;
+  }
+}
+```
+
+This will have performance impacts for the loader, but can also leave the workspace in an inconsistent state. For the scanning instrument design the workspace should be initialised all at once, via a `DetectorInfo` factory, which will ensure the loader creates a workspace in a consistent state.
```
OwenArnold: You might want to make a point of saying that the direct analogue of the problem you provided the code for above is the population of information relevant to the geometry caching `DetectorInfo` layer, such as `setPosition`. Your example above is good, and you should keep it, but it relates to the `Histogram` aspect of the workspace rather than anything that would go in the `DetectorInfo` and that needs to be made clear to avoid confusion.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -71,6 +84,27 @@ The construction of `DetectorInfo` should be done in one call, and no more time
 
 The only exception to this will be when using live data, where workspaces are built up in parts. In this case `DetectorInfo` should have a method to take another `DetectorInfo` object, and deal correctly with the shifting the time indices. Both `DetectorInfo` objects will be in a valid state, but there would still need to be checks to ensure the time indices do not overlap between the two `DetectorInfo` objects.
 
+The construction of `SpectrumInfo` should be done in a similar way. This needs to take a vector of `std::set<std::pair<detid_t, size_t>>` objects for step scans. The `StepScanInfo` could also have a method to return such an object, also ensuring it is in a consistent state with valid detector IDs and time indices. This would create an extra field in `StepScanInfo`:
+* `std::vector<std::set<std::pair<detid_t, size_t>>>`
+
+### Instrument Access
+
+Algorithms should be responsible for getting detector positions from the correct place, the `DetectorInfo`/`SpectrumInfo` class as opposed to the static instrument. However, there are potential dangers where assumptions are made about an instrument not being able to move.
```
OwenArnold: @ianbush  You might want to follow up by saying something along the lines that an `Algorithm` developer could access the `Instrument` through the caching layers and get the position from that.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -71,6 +84,27 @@ The construction of `DetectorInfo` should be done in one call, and no more time
...
+
+Moving detectors should be able to be marked as such in the instrument definition, for example `<isMovingDetector = "true" />`. The corresponding method should be added to the `IDetector` class in Mantid, for example `bool isMovingDetector() const;`.
```
OwenArnold: I'm not sure about that. You could of course always have a "scanning" instrument that's been run in a static mode for an experiment. Secondly users should always be going through the caching layers, and we should strongly discourage the `IDetector` virtual method calls and checks for performance reasons where possible. Thirdly this would seem to be a runtime check, and I'm not sure that's really what we want. For `Instrument 2.0 prototype`, I changed the method name to be `startPosition`, `startRotation` to avoid confusion.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -71,6 +84,27 @@ The construction of `DetectorInfo` should be done in one call, and no more time
...
+
+On a moving instrument access to the instrument in its default position is meaningless. Any detectors that are set to be moving should throw for direct access via the static instrument for properties such as position. This does not imply that the instrument must always be performing a step scan, as it could still do a run while in a single position. It does imply that an offset always has to be given for the detectors.
+
```
OwenArnold: @ianbush I'm not sure that this solution is the neatest. I don't think this fits with a runtime check and there will be a performance hit. Also, I think the point is that all geometry access (or at least the vast majority) even for the non-scanning cases goes through `DetectorInfo`/`SpectrumInfo`.

In terms of this design document maybe we should leave this as an somewhat open question, but say something to the effect that a runtime check will not be performed but in the eventual solution the API will make it hard to get geometry directly from the instrument itself?

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -71,6 +84,27 @@ The construction of `DetectorInfo` should be done in one call, and no more time
...
+### Saving
+
+The information relating to step scans needs loading and saving in the `LoadNexusProcessed` and `SaveNexusProcessed` algorithms. Currently the detector information is saved in the `instrument` NeXus group. Within the `detector` entry in `instrument` information about the location of the detectors is saved, with any moves to the base instrument applied. 
+
+The base instrument definition from the XML file is saved in `instrument_xml`. The parameter map is also saved in its entirety in the `instrument_parameters` entry.
+
+As the storage of this information is moved out from the parameter map and into `DetectorInfo` and `SpectrumInfo` within Mantid it should also be stored within a corresponding NeXus entries `detector_info` and `spectrum_info`. In the case of a step scan `detector_info` should contain an appropriate flag that a step scan is being stored. In this case some of the entries will contain an extra dimension corresponding to the step scan. For example a list of detector positions would be of size `n_detectors` normally, or `n_detectors x n_steps` for a step scan.
+
```
OwenArnold: 👍. We should kick off a discussion with the NeXus committee about this. Probably soon, as saving processed workspaces is a critical step and we will trip across it soon.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,170 @@
...
+The `timeIndex` argument is optional, so algorithms that do not care about time indexing will simply return the first/only time index. `DetectorInfo` should also contain a vector of start and end times for each time index. This will mean any information stored as a time series sample log can also be obtained for the correct step, for example the monitor counts for normalisation.
+
+`DetectorInfo` will ultimately store the information for positions and rotation, so will need to do this in a way that can be accessed by both the detector index and time index. Masking would be done at the spectrum level, if it is required for a detector for some time steps, but not all. `DetectorInfo` will also hold the start and end times for each step in the scan. The objects it will need to hold for this would be:
+* `std::vector<std::vector<Kernel::V3D>> positions`
+* `std::vector<std::vector<Kernel::V3D>> rotations`
+* `std::vector<std::vector<std:pair<Kernel::DateAndTime, Kernel::DateAndTime>>> timeRanges`
```
SimonHeybrock: I was thinking of doing this a bit like this, but end the end we will just have to see what works best, maybe leave this more open in the design document?
https://github.com/DMSC-Instrument-Data/instrument-prototype/pull/38#issuecomment-263863475

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,170 @@
...
+
+#### `SpectrumInfo`
+
+Currently [`SpectrumInfo`](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/SpectrumInfo.h) has a private `getDetector` method for finding the detector for the workspace. This uses the `getSpectrum(index)` method on `MatrixWorkspace` to return the detector IDs for the spectrum.
+
+A new method should be added, for example `std::set<std::pair<detid_t, size_t>> getTimeIndexedDetectorIDs(index)`. The time indices for each detector can be stored in `SpectrumInfo` as a `std::set<std::pair<detid_t, size_t>>` within a vector of the same size as the number of workspace spectra. This will avoid the need to change methods on `MatrixWorkspace` or `ISpectrum/Spectrum` to return the set of detector IDs and time indexes for the spectrum. This is something that will help to move away from the incremental construction of `DetectorInfo`.
+
+This can be used to access methods on `DetectorInfo`, for example `m_detectorInfo.twoTheta(detIndex, timeIndex)`.
```
SimonHeybrock: Do we have examples where such an access would be used? From what I have seen in most of the Mantid algorithms, work is done either with spectra, or with detectors, so there are probably hardly any cases where we would need such a mix of `SpectrumInfo` and `DetectorInfo`.

What you are saying here is certainly right, but it should be made clear that such complex interaction is probably rarely necessary.

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,170 @@
...
+
+#### `MatrixWorkspace` & `ISpectrum`/`Spectrum`
+
+The original methods such as `this->detectorIDs.insert(detID)` should remain for now. This allows existing algorithms, e.g. loaders, that do not need time indexing to work as before. Similarly the `getDetectorIDs(index)` method would still return detector IDs without the time indexing information.
+
+**Note:** working directly with the detector IDs is something that should be eliminated with the use of `SpectrumInfo` and `DetectorInfo`. The use of a `getDetectorIDs` should not be needed in the final Instrument 2.0. Similarly the method `getSpectrum(index)` on `MatrixWorkspace` should eventually be eliminated.
+
+#### StepScan Construction
```
SimonHeybrock: I have to think a bit about how this will work while we are still wrapping the old instrument. In particular construction order of workspace and instrument, etc.

Design/Instrument-2.0/ScanningInstruments_Design.md
SimonHeybrock: I am not sure I see the issue. Spectra have a fixed position and, e.g., the `DetectorInfo::position(size_t)` (without time index) would throw if a detector is moving. Do you have an example where these dangers would show up?

Design/Instrument-2.0/ScanningInstruments_Design.md
SimonHeybrock: I think I agree with @OwenArnold. But I guess the conclusion here depends on what I said 2 lines above.

Design/Instrument-2.0/ScanningInstruments_Design.md
SimonHeybrock: I think the difference between "default position" and actual position is an artifact from Instrument-1.0. Are we not basically eliminating the distinction with v2.0?

Design/Instrument-2.0/ScanningInstruments_Design.md
```diff
@@ -0,0 +1,170 @@
...
+
+void setPosition(const size_t index, const Kernel::V3D &position);
+void setPosition(const size_t index, const Kernel::V3D &position, const size_t timeIndex);
+void setRotation(const size_t index, const Kernel::Quat &rotation);
+void setRotation(const size_t index, const Kernel::Quat &rotation, const size_t timeIndex);
+
+void setPosition(const Geometry::IComponent &comp, const Kernel::V3D &pos);
+void setPosition(const Geometry::IComponent &comp, const Kernel::V3D &pos, const size_t timeIndex);
+void setRotation(const Geometry::IComponent &comp, const Kernel::Quat &rot);
+void setRotation(const Geometry::IComponent &comp, const Kernel::Quat &rot, const size_t timeIndex);
```
SimonHeybrock: This forwards to the instrument tree, component positions would not be stored in `DetectorInfo`, so I do not know how this could be implemented?

Design/Instrument-2.0/ScanningInstruments_Design.md
OwenArnold: @SimonHeybrock we are. However, these caching layers will have to sit over the current `v1.0` instrument until the replacement is ready.

Design/Instrument-2.0/ScanningInstruments_Design.md
ianbush: What I wrote here was a bit unclear. What I meant was there are dangers in accessing the static instrument, when it is a moving instrument. Yes, `DetectorInfo::position(size_t)` should certainly throw for a moving instrument.

Design/Instrument-2.0/ScanningInstruments_Design.md
ianbush: This sounds good, I think there is a clear consensus this is not required so I will remove this part and replace it with something along the lines of all geometry access going through `DetectorInfo` and `SpectrumInfo`.

Design/Instrument-2.0/ScanningInstruments_Design.md
ianbush: I'll make clear this is a rare way of accessing it.

PR #38, opened 2017-02-20
Design of fit functions python API

mantid-roman:
Describes a python wrapper class for fit functions.

### DIFF
--- a/Design/Fit functions in python.md
+++ b/Design/Fit functions in python.md
@@ -196,7 +196,7 @@ Python function `Fit` should accept instances of `FitFunctionWrapper` as its `Fu
 
 ## Function evaluation
 
-`FitFunctionWrapper` class will implement `__call__(...)` method to evaluate the function. The method will be able to accept a variety of input types:
+`FitFunctionWrapper` class will implement `__call__(...)` method to evaluate the function. The method will be able to accept a variety of input types for its first positional argument:
  1. Workspace
  2. Numpy array
  3. List of numbers
@@ -210,6 +210,34 @@ The output type is the same as that of the input. For example:
   y_list = g([-2, -1, 0, 1, 2])
   out_ws = g(ws)
 ```
+
+If the `__call__` method has the signature `__call__(self, x, *params)` then the function wrappers could be used directly with `scipy.optimize` fitting as the following example demonstrates:
+```
+import scipy.optimize
+
+class Linear:
+    
+    def __init__(self):
+        self.fun = FunctionFactory.createFunction('LinearBackground')
+        
+    def __call__(self, x, *params):
+        for i in range(len(params)):
+            self.fun.setParameter(i, params[i])
+        y = x[:]
+        ws = CreateWorkspace(x, y)
+        out = EvaluateFunction(str(self.fun), ws, OutputWorkspace='out')
+        return out.readY(1)
+        
+        
+fun = Linear()
+x = [1, 2, 3]
+y = [3, 2, 1]
+sigma = [0.1, 0.2, 0.3]
+popt, pcov = scipy.optimize.curve_fit(fun, x, y, p0=[0, 0], sigma=sigma)
+print popt
+print fun(x)
+```
+The actual implementation of `__call__` will take into account any ties that are set on the parmeters.
 
 ## Plotting
 
--- a/Design/Fit functions in python.md
+++ b/Design/Fit functions in python.md
@@ -44,14 +44,37 @@ In both ways of constructing composites the functions passed as arguments should
 
 ### Multi-domain functions
 
-A [`MultiDomainFunction`](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/MultiDomainFunction.h) needs a custom constructor which sets domain indices to its members.
+A [`MultiDomainFunction`](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/MultiDomainFunction.h) is a kind of a `CompositeFunction` that can do calculations on multiple domains (e.g. multiple spectra) that are numbered and can be identified by *domain indices*. Each member function is applied to one or more domains and can have one or more domain indices associated with it. Therefore `MultiDomainFunction` needs a custom constructor to sets domain indices to its members.
 
-Usage:
-```
-  md_fun = MultiDomainFunction(Gaussian(PeakCentre=1, Sigma=0.1), Gaussian(PeakCentre=1, Sigma=0.2), ...)
-```
+The only type of mapping that has been used in Mantid is one-to-one: one function - one domain. 'Global' fitting is achieved by tying some of the parameters together. Here it is proposed to restrict (to start with) the python interface to this use case only.
 
-To implement the constructor methods of C++ class `MultiDomainFunction` it needs to be exported to python. In particular the `setDomainIndex` method.
+So a construction of a `MultiDomainFunction` should be like this:
+```
+  md_fun = MultiDomainFunction(Gaussian(PeakCentre=1, Sigma=0.1), Gaussian(PeakCentre=1, Sigma=0.2), ..., global=['Height'])
+```
+Each function passed to the constructor will be applied to its own domain making the number of domains equal to the number of member functions. All `Height` parameters will be tied and therefore global and all the others will be local to each domain.
+
+To implement the constructor the methods of C++ class `MultiDomainFunction` need to be exported to python. In particular the `setDomainIndex` method. Here is a (very simplified) way to implement the constructor:
+```
+class MultiDomainFunctionWrapper(CompositeFunctionWrapper):
+
+    def __init__(self, *args, **kwargs):
+        # Store a reference to the C++ instance of the functon
+        self.func = FunctionFactory.createFunction('MultiDomainFunction')
+        
+        # Populate it with member functions
+        for arg in args:
+            # Each arg must be a wrapper holding a ref to its function instance
+            self.func.add(arg.func)
+        
+        # Tie the global parameters
+        self.tieAll(*kwargs['global'])
+        
+        # Set domain indices: 1-to-1
+        for i in range(len(args)):
+            self.setDomainIndex(i, i)
+
+```
 
 ## Updating functions
 
--- a/Design/Fit functions in python.md
+++ b/Design/Fit functions in python.md
@@ -23,7 +23,7 @@ The implementation must check each keyword argument if it's a parameter or an at
 
 ### Construction of composite functions
 
-Composite functions can be created in a similar way, by callig a constructor. It should be able to accept the same types of arguments as a simple function but also it must have the `functions` keyword argument that takes a list of member functions. For example:
+Composite functions can be created in a similar way, by calling a constructor. It should be able to accept the same types of arguments as a simple function but also it must have the `functions` keyword argument that takes a list of member functions. For example:
 ```
   c = CompositeFunction(functions=[Gaussian(PeakCentre=1), Gaussian(PeakCentre=2)])
 ```
@@ -49,13 +49,13 @@ Usage:
   md_fun = MultiDomainFunction(Gaussian(PeakCentre=1, Sigma=0.1), Gaussian(PeakCentre=1, Sigma=0.2), ...)
 ```
 
-To implement the constructor methods of C++ class `MultiDomainFunction` need to be exported to python. In particular the `setDomainIndex` method.
+To implement the constructor methods of C++ class `MultiDomainFunction` it needs to be exported to python. In particular the `setDomainIndex` method.
 
 ## Updating functions
 
 ### Getting and setting parameters and attributes
 
-`FitFunctionWrapper` will implement `__getitem__(self, name)` and `__setitem__(self, name, value)` methods to access parameters and attributes in a dictionary-like style (`name` is a string):
+`FitFunctionWrapper` will implement the `__getitem__(self, name)` and `__setitem__(self, name, value)` methods to access parameters and attributes in a dictionary-like style (`name` is a string):
 ```
   sigma = gaussian['Sigma']
   
@@ -164,7 +164,7 @@ This should constrain the named parameters in all members that have them. Member
 
 ## Fitting
 
-Python function `Fit` should accept instances of `FitFunctionWrapper` as its `Function` argument. `Fit` does't modify the input function but the returned value has the output function with the fitted parameters. The namedtuple returned by `Fit` must contain the fitted function:
+Python function `Fit` should accept instances of `FitFunctionWrapper` as its `Function` argument. `Fit` doesn't modify the input function but the returned value has the output function with the fitted parameters. The namedtuple returned by `Fit` must contain the fitted function:
 ```
   res = Fit(func, ws)
   output_func = res.Function
--- a/Design/Fit functions in python.md
+++ b/Design/Fit functions in python.md
@@ -4,7 +4,7 @@ Although fitting functions have some of the functionality exposed to python the 
 
 ## Function construction
 
-For each concrete function type (Gaussian, Lorentzian, etc) there should exist a corresponding python function with the same name automatically generated in a way similar to the algorithm functions. A function object is then constructed by calling such a function (constructor) and is an instance of `FitFunctionWrapper`. For example
+For each concrete function type (Gaussian, Lorentzian, etc) there should exist a corresponding python function in `simpleapi.fitfunctions` namespace with the same name automatically generated in a way similar to the algorithm functions. A function object is then constructed by calling such a function (constructor) and is an instance of `FitFunctionWrapper`. For example
 ```
   g = Gaussian()
 ```
@@ -62,19 +62,6 @@ To implement the constructor methods of C++ class `MultiDomainFunction` it needs
   comp_func['f1.A0'] = 0.5
 ```
 
-Parameters of members of composite functions can be accessed via their wrapper objects. For example:
-```
-  bk = LinearBackground()
-  peak = Lorentzian()
-  spectrum = bk + peak
-  ...
-  bk['A0'] = 1
-  peak['FWHM'] = 0.123
-  assert spectrum['f0.A0'] == 1
-  assert spectrum['f1.FWHM'] == 0.123
-```
-This allows python variables to serve as named references to the member functions that simplifies parameter and attribute setting.
-
 Implement `__getattr__()` and `__setattr__()` to access parameters of non-composite functions through instance attributes with the same name as the parameter.
 ```
   bk.A0 = 1
@@ -94,6 +81,16 @@ Composite wrappers will also define the `fn` attributes (where `n` is a number) 
   c.f1.f0.A0 = 2
   c[1][0]['A0'] = 2
   c['f1.f0.A0'] = 2
+```
+They will also define attributes with names equal to the names of the member functions + a suffix. The suffix is an int numbering repeating names. For example, members of function
+```
+  spectrum = LinearBackground() + Gaussian(PeakCentre=1) + Gaussian(PeakCentre=2)
+```
+can be refered to like this:
+```
+  spectrum.LinearBackground
+  spectrum.Gaussian0
+  spectrum.Gaussian1
 ```
 
 The wrappers will override `__iadd__(self, func)` and `__delitem__(self, i)` (`i` is an `int`, `func` is a `FitFunctionWrapper`) to implement adding and deleting members via `+=` and `del` operators:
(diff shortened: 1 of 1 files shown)


Design/Fit functions in python.md
```diff
@@ -0,0 +1,179 @@
...
+```
+  p = Polynomial(attributes={'n': 3}, parameters={'A0': 1, 'A1': 2, 'A3': 3, 'A4': 4})
+```
+Parameter and attribute names can also be used as the keys:
+```
+  g = Gaussian(Height=1, Sigma=0.1)
+  
+  p = Polynomial(n=3, A0=1, A1=2, A3=3, A4=4)
+```
+The implementation must check each keyword argument if it's a parameter or an attribute by calling `IFunction::hasAttribute(name)`. All passed attributes must be set before setting parameters. There is need to be a provision for setting attributes in a particular order in case it is essential for a fitting function. This can be done for example by passing a list of name/value pairs (tuples or lists) to the `attributes` keyword argument.
```
martyngigg: Is there a use case for having attributes set in a particular order at the moment? It seems restrictive and possibly error-prone for the user to have to remember something like that.

Design/Fit functions in python.md
```diff
@@ -0,0 +1,179 @@
...
+  c = LinearBackground() + Gaussian(PeakCentre=1) + Gaussian(PeakCentre=2)
+```
+`ProductFunction` can be constructed from a product of member functions:
+```
+  p = ExpDecay() * UserFunction(Formula='sin(w*x)', w=1)
+```
+
+### Multi-domain functions
+
+A [`MultiDomainFunction`](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/MultiDomainFunction.h) needs a custom constructor which sets domain indices to its members.
```
martyngigg: Can we have an example of this?

Design/Fit functions in python.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Updating functions
+
+### Getting and setting parameters and attributes
+
+`FitFunctionWrapper` will implement `__getitem__(self, name)` and `__setitem__(self, name, value)` methods to access parameters and attributes in a dictionary-like style (`name` is a string):
+```
+  sigma = gaussian['Sigma']
+  
+  comp_func['f1.A0'] = 0.5
```
martyngigg: As you say below, composite functions will be indexable so can we get rid of the need for the `f1` and have

```
comp_func[1]['A0'] = 0.5
```

instead? It feels more Pythonic this way.

I also wondered whether we should consider have the parameters (and attributes) exposed as Python attributes so you could do

```
comp_func[1].A0 = 0.5
```

Would that make sense?

Design/Fit functions in python.md
mantid-roman: In `CubicSpline` attribute `n` must be set before setting `X3`, `X4`, ... because they wouldn't exist yet.

Design/Fit functions in python.md
```diff
@@ -0,0 +1,179 @@
...
+  bk = LinearBackground()
+  peak = Lorentzian()
+  spectrum = bk + peak
+  ...
+  bk['A0'] = 1
+  peak['FWHM'] = 0.123
+  assert spectrum['f0.A0'] == 1
+  assert spectrum['f1.FWHM'] == 0.123
+```
+*Question: what to do if a function becomes a member of two or more composite functions?*
```
martyngigg: I think setting some kind of flag is possibly overkill.

The question here is what kind of behaviour is least confusing to users:

1. take a copy of the function as it's added to a composite so that all subsequent changes only affect that object or
2. leave it has you have here and be clear in the documentation that altering objects after are added to the composite will continue to affect it.

I'm tempted to say that the copy approach will be less error-prone and easier to explain.

Design/Fit functions in python.md
mantid-roman: `comp_func[1]['A0'] = 0.5` will work because `comp_func[1]` returns a function. But `'f1.A0'` is a valid parameter name and the function must be able to set a value for it.
`comp_func[1].A0 = 0.5` will also make perfect sense.

Design/Fit functions in python.md
martyngigg: Can we handle this internally somehow? I think it would be best if this type of behaviour did not become the norm.

martyngigg:
Another question is what module should we put these in. I'm a little concerned that `simpleapi` is becoming quite heavy.

Design/Fit functions in python.md
mantid-roman: I wanted these variables to be named references to the member functions to make it easier to work with them. The alternative is to use things like `f[0][4][10].A1 = 5` which are not less error-prone.

martyngigg: I can see both sides of the argument here. Let's leave it as and say that the composite just holds a reference and make this clear in the documentation.

mantid-roman:
The `FitFunctionWrapper` class can live in its own module and `simpleapi` can import all public names into its namespace.

Design/Fit functions in python.md
```diff
@@ -66,10 +66,18 @@ Parameters of members of composite functions can be accessed via their wrapper o
...
 *Question: what to do if a function becomes a member of two or more composite functions?*
   1. *Ignore and let the user be responsible for it*
   2. *Set a flag on `FtFunctionWrapper` and prevent subsequent attempts to add it*
 
+Implement `__getattr__()` and `__setattr__()` to access parameters of non-composite functions through instance attributes with the same name as the parameter.
+```
+  bk.A0 = 1
+```
+
+*Question: should composite functions have attributes like `c.f0`, `c.f1`, ...?*
```
martyngigg: Yes let's add those to composites too. It might make it easier to debug the ties/constraints if you can use similar syntax directly on the command line.

Design/Fit functions in python.md
```diff
@@ -0,0 +1,187 @@
...
+Tying to a constant can also be done with the `fix` method.
+```
+  func.fix('A0')             # Fix to the current value
+  func.fix(A0=2.0, A1=1.0)   # Fix to new values
+  func.fix({'f1.A2': 1.0, 'f2.A2': 2.0})
+```
+
+Ties are removed with the `removeTie` or `free` methods.
+```
+  func.free('f1.Sigma', 'f3.Sigma')
```
martyngigg: Is there a need for 2 functions with exactly the same purpose?

Design/Fit functions in python.md
```diff
@@ -0,0 +1,187 @@
...
+  func.removeTie('f1.Sigma', 'f3.Sigma')
+```
+
+Composite functions should be able to set ties on all members with a single call:
+```
+  c.tieAll('Sigma')
```
martyngigg: What value would this tie `Sigma` to?

Design/Fit functions in python.md
```diff
@@ -0,0 +1,187 @@
...
+```
+  c.constrainAll('Height > 1', 'Sigma > 0')
+```
+This should constrain the named parameters in all members that have them. Members that don't have these parameters should be skipped.
+
+## Fitting
+
+Python function `Fit` should accept instances of `FitFunctionWrapper` as its `Function` argument. `Fit` does't modify the input function but the returned value has the output function with the fitted parameters. Instead of a tuple `Fit` will return an instance of class `FitOutput` that contains all outputs from the algorithm and gives access to them both via the index operator and python read-only properties:
+```
+  res = Fit(func, ws)
```
martyngigg: I don't think we can do this in a backwards compatible way can we?

Design/Fit functions in python.md
```diff
@@ -0,0 +1,187 @@
...
+
+Composite functions should be able to set constraints on all members with a single call:
+```
+  c.constrainAll('Height > 1', 'Sigma > 0')
+```
+This should constrain the named parameters in all members that have them. Members that don't have these parameters should be skipped.
+
+## Fitting
+
+Python function `Fit` should accept instances of `FitFunctionWrapper` as its `Function` argument. `Fit` does't modify the input function but the returned value has the output function with the fitted parameters. Instead of a tuple `Fit` will return an instance of class `FitOutput` that contains all outputs from the algorithm and gives access to them both via the index operator and python read-only properties:
```
martyngigg: How do we pass this Python object to Fit. Can you include a basic class description of the layout of the `FitFunctionWrapper` class?

Design/Fit functions in python.md
mantid-roman: `CubicSpline` can be changed by combining all `X`s into a single vector attribute. Then there is no need for that order. But here is a more evil example: `CrystalFieldMultiSpectrum`. It has attributes that through which you define the types of the member functions. When you set those attributes the parameters and attributes of the new member become parameters and attributes of this instance of `CrystalFieldMultiSpectrum`.

At the moment I cannot see a simple way of doing this internally and in a generic way. What I am suggesting is just a simple mechanism that can be used when needed.

On the other hand you can pass only the independent attributes to the constructor and set the others later in any order you like.

Design/Fit functions in python.md
mantid-roman: Maybe there is no need for it. I thought it would be symmetrical: `tie` - `removeTie` (or untie?), `fix` - `free`.

Design/Fit functions in python.md
mantid-roman: It would tie all `Sigma`s together as the following example shows.

Design/Fit functions in python.md
mantid-roman: If `FitOutput` has the `__getitem__` method that returns the outputs in the same order they are now I think we can. `__str__` can also be implemented to produce the same string.

Design/Fit functions in python.md
mantid-roman: The wrapper can have its `__str__` method return the function initialisation string. Then Fit just will have to wrap its function argument into `str()` and it will work. Or it can ask its function argument if it's an instance on `FitFunctionWrapper` then ask it to create the initialisation string.

Design/Fit functions in python.md
mantid-roman: `__iter__` can be implemented but I think it's too far.

martyngigg: We have recently started returning a [`namedtuple`](https://docs.python.org/2/library/collections.html#collections.namedtuple) from all Python algorithm functions (it will be mentioned at the next review). Will this already cover this?

Design/Fit functions in python.md
martyngigg: Is there a need to separate `tie` & `fix`? How would we explain to users when to use one over the other?

Design/Fit functions in python.md
martyngigg: Can we avoid going via strings and use the function objects directly?

Design/Fit functions in python.md
mantid-roman: `fix` can be called like this: `f.fix('A0')` and it will fix the `A0` parameter to its current value whatever it is. Function `f` can be the output of `Fit` and `A0` has changed and you want to re-run the fit with `A0` fixed now. Fixing/freeing of parameters is a usual practice in curve fitting. People who has an experience in it should find those to be familiar terms (I would expect). For those who are new to fitting we can explain that a fixed parameter doesn't change after a fit while the value of a tied parameter depends on the values of other parameters.

Design/Fit functions in python.md
mantid-roman: Ideally yes. But I think FunctionProperty isn't working with python now. You surely know more about it than I do.

If we do it we need to make sure that cloning works for all functions. I am not sure that it does.

Design/Fit functions in python.md
mantid-roman: I think it will. I didn't know that type existed.

Design/Fit functions in python.md
martyngigg: Okay, I thought it was but I'm probably wrong. At least we can handle this all internally in the future so it won't affect user code. We can have this as a future improvement.

Design/Fit functions in python.md
martyngigg: We have the option of doing what we do for the algorithms and have an exception to the generic wrapper?

Design/Fit functions in python.md
martyngigg: Okay. Can I suggest that we limit the use of `fix` to simply fixing the value to the current value i.e just `f.fix('A0')` and all other behaviour goes through `tie`?

mantid-roman: Agree.

Design/Fit functions in python.md
```diff
@@ -0,0 +1,187 @@
...
+```
+  res = Fit(func, ws)
+  output_func = res.function
+  chi2 = res.chiSquared
+  ...
+```
+
+## Function evaluation
+
+`FitFunctionWrapper` class will implement `__call__(...)` method to evaluate the function. The method will be able to accept a variety of input types:
```
martyngigg: Can you provide a sample of what you would expect the user code to look like, please.

martyngigg:
@mantid-roman I think this is nearly there. Can you go back an update the sections to address some of the comments so that the document can be read with out them.

Design/Fit functions in python.md
```diff
@@ -0,0 +1,198 @@
...
+
+Python function `Fit` should accept instances of `FitFunctionWrapper` as its `Function` argument. `Fit` does't modify the input function but the returned value has the output function with the fitted parameters. The namedtuple returned by `Fit` must contain the fitted function:
+```
+  res = Fit(func, ws)
+  output_func = res.Function
+```
+
+## Function evaluation
+
+`FitFunctionWrapper` class will implement `__call__(...)` method to evaluate the function. The method will be able to accept a variety of input types:
```
jmborr: regarding plotting, having a python method is more intuitive to me than an external function.
```
f=Lorentzian()
f.plot(*args, **args) # workspace, workspaceindex, startX, endX
```
Sometimes the workspace and workspaceIndex have been already passed with the SetWorkspace method.
```
f=TeixeiraWaterSQE()
f.setMatrixWorkspace(workspace, workspaceIndex) # determines attribute Q
f.plot(startX, endX)
```

Design/Fit functions in python.md
```diff
@@ -0,0 +1,198 @@
...
+```
+
+## Plotting
+
+Python function `plotFunction` will plot one or more fitting functions in MantidPlot. It will have the arguments:
+ 1. first positional argument - the function(s) to plot
+ 2. `workspace` - optional workspace on which x-values to evaluate the function
+ 3. `workspaceIndex` - optional index of a spectrum to get the x-values from
+ 4. `startX` - optional if `workspace` is given, mandatory otherwise. The start of the plot region.
+ 5. `endX` - optional if `workspace` is given, mandatory otherwise. The end of the plot region.
```
jmborr: regarding plotting, having a python method is more intuitive to me than an external function.
```
f=Lorentzian()
f.plot(*args, **args) # workspace, workspaceindex, startX, endX
```
Sometimes the workspace and workspaceIndex have been already passed with the SetWorkspace method.
```
f=TeixeiraWaterSQE()
f.setMatrixWorkspace(workspace, workspaceIndex) # determines attribute Q
f.plot(startX, endX)
```
or you could have both :)
```
def plot(self, *args, **kwargs):
  plotFunction(self, *args, **kwargs)
```

martyngigg: @jmborr I think we will probably end up with both similar to the method `matplotlib.Axes.plot` and function `matplotlib.pyplot.plot`

Design/Fit functions in python.md
```diff
@@ -0,0 +1,198 @@
...
+  spectrum += Lorentzian(PeakCentre=3)
+  del spectrum[0]
+```
+
+Implement `__len__(self)` to return the number of member functions.
+```
+  n_peaks = len(spectrum)
+```
+
+### Setting ties
```
peterfpeterson: names should be `fix` and `unfix`, `tie` and `untie`

Design/Fit functions in python.md
```diff
@@ -0,0 +1,198 @@
+# Working with fit functions in python
+
+Although fitting functions have some of the functionality exposed to python the use of it is very limited and functions are constructed via strings. Values of the optimised parameters get extracted from `TableWorkspace`s output from the Fit algorithm. Both function construction and extraction of the results are cumbersome and error prone. This document describes a design of a python fitting API aiming to overcome these problems. The main idea of the solution proposed here is to create and manipulate a C++ [`IFunction`](https://github.com/mantidproject/mantid/blob/master/Framework/API/inc/MantidAPI/IFunction.h) object from python directly via a thin wrapper python classes. `FitFunctionWrapper` will be the base class functionality common for all fit functions. Composite functions (`CompositeFunction`, `ProductFunction`, `Convolution`, `MultiDomainFunction`) will have their own wrappers inheriting from `FitFunctionWrapper`. If a fit function exibits a special behaviour a custom wrapper can be defined.
```
peterfpeterson: Explicitly mention that the fit functions will appear in `simpleapi.fitfunction` namespace

Design/Fit functions in python.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+Implement `__getattr__()` and `__setattr__()` to access parameters of non-composite functions through instance attributes with the same name as the parameter.
+```
+  bk.A0 = 1
+```
+
+### Managing members of composite functions
+
+Wrappers for composite functions will implement `__getitem__(self, i)` method to access members of a composite function in a list-like style (`i` is an `int`):
```
martyngigg: Can we add something about allowing iteration in a `for func in comp:` style?

Design/Fit functions in python.md
```diff
@@ -0,0 +1,199 @@
...
+Composite functions should be able to set constraints on all members with a single call:
+```
+  c.constrainAll('Height > 1', 'Sigma > 0')
+```
+This should constrain the named parameters in all members that have them. Members that don't have these parameters should be skipped.
+
+## Fitting
+
+Python function `Fit` should accept instances of `FitFunctionWrapper` as its `Function` argument. `Fit` doesn't modify the input function but the returned value has the output function with the fitted parameters.
+
```
mantid-roman: I think we still need to process the output here: `out.Function` is a `FunctionProperty`, `out.Function.value` is an `IFunction`, but we need a `FitFunctionWrapper` or the user should wrap it.

peterfpeterson:
Do we want to require that the fit functions are compatible with [`scipy.optimize`](https://docs.scipy.org/doc/scipy/reference/tutorial/optimize.html)? I've only got experience with using [`curve_fit`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.curve_fit.html), but I'm suspicious it would be good to be able to use someone else's abstraction to generic fitting algorithms.

mantid-roman:
`scipy.optimize` doesn't require the fit function itself but a cost function. It needs the function and the data to fit to. It can be done on top of this API.

jmborr:
As Roman said, `scipy.optimize.minimize` only requires the Chi-square and a numpy-vector with size equal to the number of active fit parameters. It doesn't need to know about the specifics of the fit function or the experimental data.

martyngigg:
The document suggests overriding `__call__` to evaluate the function with data and parameters so this would lead to it being quite simple to write a thin wrapper function to bind the two (you could do it with a lambda). As a start we could just include documentation on how to do this?

mantid-roman:
I am not sure how thin this wrapper will be. It will need to extract data from workspaces, calculate weights, cost function, apply ties and constraints, select free parameters - things that `Fit` does.

martyngigg:
@mantid-roman I think this is more about just using the function itself easily if for example we already it coded up and for some reason there is a desire to use the `scipy` optimize package. I think have the function callable as you have suggested basically gets you most of the way.

mantid-roman:
I can give an example of how I would do it

Design/Fit functions in python.md
```diff
@@ -211,6 +211,34 @@ The output type is the same as that of the input. For example:
...
+        
+fun = Linear()
+x = [1, 2, 3]
+y = [3, 2, 1]
+sigma = [0.1, 0.2, 0.3]
+popt, pcov = scipy.optimize.curve_fit(fun, x, y, p0=[0, 0], sigma=sigma)
+print popt
+print fun(x)
+```
+The actual implementation of `__call__` will take into account any ties that are set on the parmeters.
```
martyngigg: :+1:

peterfpeterson:
Approved by TSC on [2017-03-14](https://github.com/mantidproject/documents/blob/master/Project-Management/TechnicalSteeringCommittee/meetings/2017/TSC-meeting-2017-03-14.md)

PR #42, opened 2017-05-05
Workspace Property With Index Design Document

LamarMoore:
This document proposes a design for a new property type which combines workspace and index specification.

### DIFF
--- a/Design/WorkspacePropertyWithIndex.md
+++ b/Design/WorkspacePropertyWithIndex.md
@@ -45,7 +45,7 @@ The points are taken from the original `IndexInfo` design document linked in the
 - Extend the `WorkspaceProperty` type.
 - Own an `ArrayProperty` which can make use of `ListPropertyWidget` for capturing list data from the GUI.
 - Add a set of flags which control allowed index types (for use in the constructor).
-- Use a PropertyWithValue<std::string> to capture the selected index type.
+- Use a `PropertyWithValue<std::string>` to capture the selected index type.
 
 ### Details
 
@@ -84,25 +84,20 @@ declareProperty(make_unique<WorkspacePropertyWithIndex<MatrixWorkspace>>("InputW
 
 *C++*:
 
-Setting the property manually:
+Setting the property manually (tuple is created internally):
 
 ```cpp
 setProperty("InputWorkspaceWithIndex", ws, std::vector<SpectrumNumber>{1, 2, 3, 4});
 //or
 setProperty("InputWorkspaceWithIndex", ws, IndexType::SpectrumNumber, "1:33,42");
-//or
-setProperty("InputWorkspaceWithIndex", std::make_pair<MatrixWorkspace_sptr, std::vector<SpectrumNumber>>(ws, {1, 2, 3, 4}));
-//or 
-setProperty("InputWorkspaceWithIndex", std::make_tuple<MatrixWorkspace_sptr, IndexType, std::string>(ws, IndexType::SpectrumNumber, "1:33,42"));
 ```
 
 Accessing the property:
 
 ```cpp
-//C++11
 MatrixWorkspace_const_sptr inputWS;
 IndexSet indices;
-std::tie(inputWS, indices) = std::pair<MatrixWorkspace_const_sptr, SpectrumIndexSet>(getProperty("InputWorkspaceWithIndex"));
+std::tie(inputWS, indices) = getProperty("InputWorkspaceWithIndex");//returns tuple
 ```
 
 *Python*:
@@ -112,6 +107,9 @@ The simplest scenario for a call to an algorithm could be:
 ChangePulsetime(TimeOffset=10, InputWorkspaceWithIndex=("ws", SpectrumNumber, [1:33, 42]))
 # or 
 ChangePulsetime(TimeOffset=10, InputWorkspaceWithIndex=("ws", SpectrumNumber, "1:33, 42"))
+
+# optionally no indices. Will just assume all spectra 
+ChangePulsetime(TimeOffset=10, InputWorkspaceWithIndex=("ws", SpectrumNumber))
 ```
 Creating special handling for Index types could allow for:
 ```python
@@ -125,6 +123,7 @@ Property Access would then take the form:
 ```python
 inputWS, indices = getProperty("InputWorkspaceWithIndex")
 ```
+
 ### Development Stages and Roll-Out
 
 - The first release of this type will only include support for `IndexType::SpectrumNumber` and `IndexType::WorkspaceIndex`. There `DetectorID` translation, which is rarely used, is tricky and is not currently supported within IndexInfo, another translator specific to detector ids must be developed before this mode can be supported.
--- a/Design/WorkspacePropertyWithIndex.md
+++ b/Design/WorkspacePropertyWithIndex.md
@@ -91,9 +91,9 @@ setProperty("InputWorkspaceWithIndex", ws, std::vector<SpectrumNumber>{1, 2, 3, 
 //or
 setProperty("InputWorkspaceWithIndex", ws, IndexType::SpectrumNumber, "1:33,42");
 //or
-setProperty("InputWorkspaceWithIndex", std::tie(ws, std::vector<SpectrumNumber>{1, 2, 3, 4}));
+setProperty("InputWorkspaceWithIndex", std::make_pair<MatrixWorkspace_sptr, std::vector<SpectrumNumber>>(ws, {1, 2, 3, 4}));
 //or 
-setProperty("InputWorkspaceWithIndex", std::tie(ws, IndexType::SpectrumNumber, "1:33,42"));
+setProperty("InputWorkspaceWithIndex", std::make_tuple<MatrixWorkspace_sptr, IndexType, std::string>(ws, IndexType::SpectrumNumber, "1:33,42"));
 ```
 
 Accessing the property:
--- a/Design/WorkspacePropertyWithIndex.md
+++ b/Design/WorkspacePropertyWithIndex.md
@@ -71,14 +71,16 @@ The points are taken from the original `IndexInfo` design document linked in the
 Declaring the property in C++:
 ```cpp
 //Property name may be fixed?
-declareProperty(make_unique<WorkspacePropertyWithIndex<MatrixWorkspace>>("InputWorkspaceWithIndex", IndexType::SpectrumNumber|IndexType::WorkspaceIndex));
+declareProperty(WorkspacePropertyWithIndex("InputWorkspace", IndexType::SpectrumNumber|IndexType::WorkspaceIndex));
+//IndexType::WorkspaceIndex is the defaul
+declareProperty(WorkspacePropertyWithIndex("InputWorkspace"));
 ```  
 In order to reduce typing we could include builder functions which allow for more concise property declaration:
 ```cpp
 //Builder method returns a unique_ptr?
 declareProperty(CreateMatrixWorkspaceWithSpectrumNumber("InputWorkspaceWithIndex");
-//or
-declareProperty(CreateEventWorkspaceWithSpectrumNumber("InputWorkspaceWithIndex");
+//or alternatively
+declareProperty(CreateWithSpectrumNumber<MatrixWorkspace>("InputWorkspaceWithIndex");
 ```
 
 **Algorithm Dialog Box (GUI)**:
@@ -110,17 +112,23 @@ std::tie(inputWS, indices) = getProperty("InputWorkspaceWithIndex");//returns tu
 *Python*:
 
 
-In order to reduce the load on script authors, we could split keyword arguments and create types which capture each index type:
+In order to reduce the load on script authors, we could split keyword arguments and create types which capture each index type. This expanding arguments allows users to understand the autocomplete in the scripting editor more easily.:
 ```python
 # The index types could accept a range as a string or
 # a numpy array which contains the values.
 ChangePulseTime(TimeOffset=10, InputWorkspace=ws, Spectra=("1:33") )
-ChangePulseTime(TimeOffset=10, InputWorkspace=ws, Indices=("1:33") )
-ChangePulseTime(TimeOffset=10, InputWorkspace=ws, IDs=("1:33") )
+ChangePulseTime(TimeOffset=10, InputWorkspace=ws, Indices=("0:32") )
+ChangePulseTime(TimeOffset=10, InputWorkspace=ws, IDs=("1001:1101") )
+```
+For multiple properties of this type in an algorithm, we could enforce a policy of pre-pending the property name on the extra keyword:
+```python
+# The index types could accept a range as a string or
+# a numpy array which contains the values.
+ChangePulseTime(TimeOffset=10, Workspace1=ws, Workspace2=ws, Workspace1Spectra=("1:33"), Workspace2Indices=("0:32"))
 ```
 Property Access would then take the form:
 ```python
-inputWS, indices = getProperty("InputWorkspaceWithIndex")
+inputWS, indices = getProperty("InputWorkspace")
 ```
 
 ### Development Stages and Roll-Out
--- a/Design/WorkspacePropertyWithIndex.md
+++ b/Design/WorkspacePropertyWithIndex.md
@@ -1,0 +1,95 @@
+# WorkspacePropertyWithIndex Type
+
+This document relates to an algorithm property type which can be used to specify a workspace and desired input index type(s). Index translation will will be handled by the new `IndexInfo` (design [here](https://github.com/mantidproject/documents/blob/master/Design/spectrum_number_and_workspace_index_abstraction.md)) object internally. Badly unit tested code and inconsistencies in validating user indices would be absorbed by this type. This design complements `IndexInfo` and reinforces some of the design concepts. **Note**:The property would then return the workspace and workspace indices (after automatic translation).
+## Motivation
+The `IndexInfo` object recently included in the Mantid framework introduces a strict policy for the definition of and translations between index types. However, this has not yet been reflected at the algorithm property level. There is still user-facing confusion among index types. Algorithms inconsistently capture and use index types without a clear mechanism for defining them e.g spectrum number, spectrum index, workspace index, workspace id, detector index, detector id etc. This is compounded by the soon to be implemented MPI build of Mantid. Indices captured by algorithms must be translated in accordance with data partitioning across MPI ranks. The mechanism in place for this translation exists within *IndexInfo* `IndexInfo::makeIndexSet` but developers must manually make this translation upon obtaining a workspace property.
+
+```cpp
+MatrixWorkspace_sptr inputWs = getProperty("InputWorkspace");
+std::vector<int> indices = getProperty("WorkspaceIndices");
+auto indexSet = inputWs->indexInfo().makeIndexSet(indices);
+```
+ A developer could easily miss this step which would have undefined results in an MPI situation. It would be more sensible to abstract this away from the developer. The `WorkspacePropertyWithIndex` type will obviate the need to perform this translation by doing this internally and returning the correct index set for use by the algorithm. In summary, this property type presents the following benefits:
+
+- Maintaining consistency in how index types are presented and used.
+- Reduce the likelihood of errors due to incorrect use of `IndexInfo`.
+- Hide the details of MPI from the developer.
+
+## Current Situation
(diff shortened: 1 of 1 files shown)


Design/WorkspacePropertyWithIndex.md
```diff
@@ -0,0 +1,95 @@
...
+
+```cpp
+MatrixWorkspace_sptr inputWs = getProperty("InputWorkspace");
+std::vector<int> indices = getProperty("WorkspaceIndices");
+auto indexSet = inputWs->indexInfo().makeIndexSet(indices);
+```
+ A developer could easily miss this step which would have undefined results in an MPI situation. It would be more sensible to abstract this away from the developer. The `WorkspacePropertyWithIndex` type will obviate the need to perform this translation by doing this internally and returning the correct index set for use by the algorithm. In summary, this property type presents the following benefits:
+
+- Maintaining consistency in how index types are presented and used.
```
SimonHeybrock: Maybe add (either separately, or as subitems)
- consistent and unit-tested translation without code duplication
- consistent and unit-tested validation without code duplication
- consistent and simple way of defining index properties without code duplication
- consistent user interface

Edit: I now see that this is mostly discussed on the next section.

Design/WorkspacePropertyWithIndex.md
```diff
@@ -0,0 +1,95 @@
...
+2. Users will be restricted to entering only one of the allowed index types at the GUI level. 
+3. Validation will occur in `IndexInfo::makeIndexSet`(which throws for bad indices), not in every algorithm.
+4. The `ArrayProperty` will be exposed e.g `WorkspacePropertyWithIndex::getArrayProperty` in order to make use of `ListPropertyWidget`.
+5. Overload function call operator to return `std::pair`.
+6. `IndexInfo` will be used to translate the indices at the point of access i.e. `WorkspacePropertyWithIndex::operator()()`.
+7. There must be a few modifications to the GUI generation code in order to support this type:
+	-  `PropertyWidgetFactory` may need to handle `WorkspacePropertyWithIndex` specifically.
+	-  May need to introduce a concept of a group of property widgets.
+
+<sup>1</sup><sub>The distinction is made here between local (a subset of all data partitioned on an MPI Rank) and global (entire data set).</sub>
```
SimonHeybrock: Maybe clarify that this distinction will never be visible on the UI side, i.e., the user facing API or GUI will *always* work with global indices.

Design/WorkspacePropertyWithIndex.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Example usage
+
+**Property Declaration**:
+
+```cpp
+//Property name is fixed
+declareProperty(make_unique<WorkspacePropertyWithIndex>(IndexType::SpectrumNumber|IndexType::WorkspaceIndex));
```
SimonHeybrock: Maybe clarify that you also need to pass the workspace type (`WorkspaceProperty` is a template)?

Design/WorkspacePropertyWithIndex.md
```diff
@@ -0,0 +1,95 @@
...
+	-  May need to introduce a concept of a group of property widgets.
+
+<sup>1</sup><sub>The distinction is made here between local (a subset of all data partitioned on an MPI Rank) and global (entire data set).</sub>
+
+### Example usage
+
+**Property Declaration**:
+
+```cpp
+//Property name is fixed
```
SimonHeybrock: I know I suggested this, but I think this is up for discussion, i.e., is there always a sensible and fixed name (`InputWorkspace`)?

Design/WorkspacePropertyWithIndex.md
```diff
@@ -0,0 +1,95 @@
...
+declareProperty(make_unique<WorkspacePropertyWithIndex>(IndexType::SpectrumNumber|IndexType::WorkspaceIndex));
+```  
+
+**Algorithm Dialog Box (GUI)**:
+
+**InputWorkspaceWithIndex** [ _________________ ]<br>
```
SimonHeybrock: I think this should just display `InputWorkspace`

Design/WorkspacePropertyWithIndex.md
```diff
@@ -0,0 +1,95 @@
...
+MatrixWorkspace_const_sptr inputWS;
+IndexSet indices;
+std::tie(inputWS, indices) = getProperty("InputWorkspaceWithIndex");
+```
+
+Python:
+
+```python
+inputWS, indices = getProperty("InputWorkspaceWithIndex")
+```
```
SimonHeybrock: Could you add a discussion of how an algorithm with such a property would be called? What would you pass to `setProperty` in C++? In Python, would there be a single argument that accepts a compount argument containing the workspace and this indices?

Even if you do not have a solution ready, putting something in here aids the discussion.

Design/WorkspacePropertyWithIndex.md
```diff
@@ -0,0 +1,95 @@
+# WorkspacePropertyWithIndex Type
+
+This document relates to an algorithm property type which can be used to specify a workspace and desired input index type(s). Index translation will will be handled by the new `IndexInfo` (design [here](https://github.com/mantidproject/documents/blob/master/Design/spectrum_number_and_workspace_index_abstraction.md)) object internally. Badly unit tested code and inconsistencies in validating user indices would be absorbed by this type. This design complements `IndexInfo` and reinforces some of the design concepts. **Note**:The property would then return the workspace and workspace indices (after automatic translation).
```
SimonHeybrock: Consider adding another note: This is about index properties that *are used to obtain a workspace index*. This property does *not* cater for applications that directly need, e.g., detector IDs.

Design/WorkspacePropertyWithIndex.md
```diff
@@ -56,36 +62,66 @@ The points are taken from the original `IndexInfo` design document linked in the
...
+```
+
+Accessing the property:
 
 ```cpp
 //C++11
 MatrixWorkspace_const_sptr inputWS;
 IndexSet indices;
-std::tie(inputWS, indices) = getProperty("InputWorkspaceWithIndex");
+std::tie(inputWS, indices) = std::pair<MatrixWorkspace_const_sptr, SpectrumIndexSet>(getProperty("InputWorkspaceWithIndex"));
```
SimonHeybrock: Why is this needed? Can we avoid writing the types twice?

Design/WorkspacePropertyWithIndex.md
```diff
@@ -56,36 +62,66 @@ The points are taken from the original `IndexInfo` design document linked in the
...
 ```
 
-Python:
+*Python*:
 
+The simplest scenario for a call to an algorithm could be:
+```python
+ChangePulsetime(TimeOffset=10, InputWorkspaceWithIndex=("ws", SpectrumNumber, [1:33, 42]))
```
SimonHeybrock: I think I would simply name the property `InputWorkspace`. Typically, indices are optional, so we also want to be able to support `InputWorkspace=ws` (without specifying indices). Having `InputWorkspaceWithIndex=ws` would look weird.

Design/WorkspacePropertyWithIndex.md
```diff
@@ -56,36 +62,66 @@ The points are taken from the original `IndexInfo` design document linked in the
...
+setProperty("InputWorkspaceWithIndex", ws, std::vector<SpectrumNumber>{1, 2, 3, 4});
+//or
+setProperty("InputWorkspaceWithIndex", ws, IndexType::SpectrumNumber, "1:33,42");
+//or
+setProperty("InputWorkspaceWithIndex", std::tie(ws, std::vector<SpectrumNumber>{1, 2, 3, 4}));
+//or 
+setProperty("InputWorkspaceWithIndex", std::tie(ws, IndexType::SpectrumNumber, "1:33,42"));
+```
+
+Accessing the property:
```
SimonHeybrock: Maybe clarify that you are not suggesting to support *all* of the options given below, but only one of them (which one is to be determined).

Design/WorkspacePropertyWithIndex.md
LamarMoore: std::tie automatically evaluates to a tuple (although it can upack a pair). So if we use std::tuple in the implementation, I believe we can avoid writing the types twice.

Design/WorkspacePropertyWithIndex.md
LamarMoore: In the C++ case I was thinking we could support all of the options. It wouldn't be difficult to do so.

Design/WorkspacePropertyWithIndex.md
LamarMoore: Yes but then there would be confusion as to what the type exists since there will be workspace properties called `InputWorkspace` which strictly accept workspaces.

Design/WorkspacePropertyWithIndex.md
SimonHeybrock: I think supporting both strings and vectors is good, but we should choose between having separate arguments and passing tuples.

Design/WorkspacePropertyWithIndex.md
SimonHeybrock: That is also a good point, I guess this is up for discussion. Could you add a note that specifying indices is optional (also with a code example)?

Design/WorkspacePropertyWithIndex.md
LamarMoore: I guess we could accept seperate arguments and just build the tuple internally.

Design/WorkspacePropertyWithIndex.md
```diff
@@ -112,6 +107,9 @@ The simplest scenario for a call to an algorithm could be:
 ChangePulsetime(TimeOffset=10, InputWorkspaceWithIndex=("ws", SpectrumNumber, [1:33, 42]))
 # or 
 ChangePulsetime(TimeOffset=10, InputWorkspaceWithIndex=("ws", SpectrumNumber, "1:33, 42"))
+
+# optionally no indices. Will just assume all spectra 
+ChangePulsetime(TimeOffset=10, InputWorkspaceWithIndex=("ws", SpectrumNumber))
```
SimonHeybrock: In that case the argument `SpectrumNumber` should also not be there, I think? And we do not need to pass a tuple.

Design/WorkspacePropertyWithIndex.md
```diff
@@ -0,0 +1,130 @@
...
+<sup>1</sup><sub>The distinction is made here between local (a subset of all data partitioned on an MPI Rank) and global (entire data set). This will not be visible at the GUI or client API level.</sub>
+
+### Example usage
+
+**Property Declaration**:
+
+Declaring the property in C++:
+```cpp
+//Property name may be fixed?
+declareProperty(make_unique<WorkspacePropertyWithIndex<MatrixWorkspace>>("InputWorkspaceWithIndex", IndexType::SpectrumNumber|IndexType::WorkspaceIndex));
```
martyngigg: Could it be worth having some high-level builder type functions that could create the commonly used types to cut down on the overhead in typing and to aid with readability?

Design/WorkspacePropertyWithIndex.md
```diff
@@ -0,0 +1,130 @@
...
+MatrixWorkspace_const_sptr inputWS;
+IndexSet indices;
+std::tie(inputWS, indices) = getProperty("InputWorkspaceWithIndex");//returns tuple
+```
+
+*Python*:
+
+The simplest scenario for a call to an algorithm could be:
+```python
+ChangePulsetime(TimeOffset=10, InputWorkspaceWithIndex=("ws", SpectrumNumber, [1:33, 42]))
```
martyngigg: I don't think `[1:33, 42]` is valid syntax without an object to slice on.

I'm not too sure about the general form here - it's a lot for users to remember to type. I don't think there's any autocomplete help we could add here. I think this is the most difficult part though.

What about always generating the same available keyword arguments in Python and then combining them internall into the WorkspacePropertyWithIndex, e.g

```python
ChangePulseTime(TimeOffset=10, InputWorkspace=ws, Spectra=("1:33") )
ChangePulseTime(TimeOffset=10, InputWorkspace=ws, Indices=("1:33") )
ChangePulseTime(TimeOffset=10, InputWorkspace=ws, IDs=("1:33") )
```
In this case the default of leaving out the `Spectra/Indices/IDs` just applies it to the whole workspace as now? 

Is this too ambiguous for people?

Design/WorkspacePropertyWithIndex.md
SimonHeybrock: :+1: 
I think there are several aspects:
- Workspace template argument: `WorkspaceProperty` has `MatrixWorkspace` as a default, I suppose we can keep that.
- `IndexType` arguments: I think we should probably agree on a default (personally I would prefer only `WorkspaceIndex`, for performance reasons, but we can argue about that, I suppose many would argue to have also `SpectrumNumber` there by default, in addition to `WorkspaceIndex`).
- `make_unique` (not specific to this new proposal): Why do we need this for `WorkspaceProperty`? Can't we just pass the property directly?
- Property name?

Design/WorkspacePropertyWithIndex.md
SimonHeybrock: 1. Yes, leaving out the indices implies "whole workspace".
2. I think we had discussed having split keyword arguments somewhere, I don't remember where. I like it, but I have some concerns that we should think about it in the design: We do *not* want to allow specifying two types of indices (there are currently some algorithms that do that, and then merge the indices obtained from different properties into a combined set of indices). Ideally, it should simply not be possible to do so (otherwise we need extra code to check for that). One option would be to have something like

  ```python
  ChangePulseTime(TimeOffset=10, InputWorkspace=ws, IndexType="SpectrumNumbers", Indices="1:33")
  ```

  But I agree that this is longer than you suggestion, which is not nice.

Design/WorkspacePropertyWithIndex.md
LamarMoore: @martyngigg do you mean something like:
```cpp
declareProperty(CreateMatrixWorkspaceWithSpectrumNumber("InputWorkspace"));
declareProperty(CreateEventWorkspaceWithWorkspaceIndex("InputWorkspace"));
```

The builder name could infer the index type. Or are these names too long?

Design/WorkspacePropertyWithIndex.md
LamarMoore: I agree with @SimonHeybrock  I think there is a danger that users could specify multiple types of indices (we will eventually include handling for `DetectorID`). If we use a type for the IndexType, then users will only really need to remember how to define the indices if they are always accepted as a string.

```python
ChangePulseTime(TimeOffset=10, InputWorkspace=ws, IndexType=IndexType::SpectrumNum, Indices="1:33,50")
```

martyngigg: Could't we trap this in the Python layer at the point where the Python function for the algorithm is defined? If we are moving towards standardizing the property names then we could validate that only one of them has been passed in?

I think the type is a good idea at the developer level I'm just concerned that it's a lot to explain to our users who generally have limited knowledge of these types of thing.

LamarMoore: @martyngigg then throw an exception if multiple index types are defined?

Design/WorkspacePropertyWithIndex.md
martyngigg: I think as @SimonHeybrock says may we can sort this out by having `MatrixWorkspace` as a default template parameter and default `IndexType::WorkspaceIndex`. That would cover most cases and then only some places would need the explicit version.

`PropertyManager` currently managers pointers so `make_unique` just ensures the pointer has been created - I'm not sure how best to deal with that. It would be nice just to have the `PropertyManager` move construct a new object and handle that.

Design/WorkspacePropertyWithIndex.md
martyngigg: Yes, that was my first thought.

Design/WorkspacePropertyWithIndex.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+*Python*:
+
+
+In order to reduce the load on script authors, we could split keyword arguments and create types which capture each index type. This expanding arguments allows users to understand the autocomplete in the scripting editor more easily.:
+```python
+# The index types could accept a range as a string or
+# a numpy array which contains the values.
+ChangePulseTime(TimeOffset=10, InputWorkspace=ws, Spectra=("1:33") )
```
martyngigg: I agree that I think this is the best option from a user perspective as it introduces the fewest new concepts :+1:

martyngigg:
I'm happy with this document as it stands.

mantid-roman:
Have you considered having validators that check multiple properties instead? It could be a more general case and be used for example to validate `Start` and `End` properties, various mins and maxs, etc.

martyngigg:
@mantid-roman That's an interesting point. We kind of have that with `validateInputs` don't we?

In this case I think Lamar's proposal goes further on making it harder to do the wrong thing when retrieving the workspace/range and removes a lot of boiler plate code on many algorithms. For future MPI versions this type of checking code really needs to be centralised.

Design/WorkspacePropertyWithIndex.md
```diff
@@ -0,0 +1,137 @@
...
+- Add a set of flags which control allowed index types (for use in the constructor).
+- Use a `PropertyWithValue<std::string>` to capture the selected index type.
+
+### Details
+
+1. The `WorkspacePropertyWithIndex` type constructor will accept the following flags (or combinations of flags):
+	- `IndexType::SpectrumNumber` for list of global<sup>1</sup> spectrum numbers typically from 1-N.
+	- `IndexType::WorkspaceIndex` for list of global workspace indices typically from 0-N.
+	- `IndexType::DetectorID` for list of detector ids within the instrument.
+2. Users will be restricted to entering only one of the allowed index types at the GUI level.
```
SimonHeybrock: Yes this is referring to the algorithm dialog. How are `CustomInterfaces` handling this? Are they simply forwarding to an algorithm?

martyngigg: My suggestion here would be that we implement a widget type specifically for the WorkspacePropertyWithIndex, see [PropertyWidgetFactory](/mantidproject/mantid/blob/master/MantidQt/API/inc/MantidQtAPI/PropertyWidgetFactory.h)

DTasev: I would think that they would forward the input to the algorithm property, after reading from something like a `QTextEdit`

LamarMoore:
@martyngigg @mantid-roman The checking is centralized in `indexInfo`, the `WorkspacePropertyWithIndex` forces this central check. Person's writing an algorithm will never have to partition data manually. This will be handled by the selected execution mode of the algorithm and the partition mode of the `SpectrumNumberTranslator`.

Design/WorkspacePropertyWithIndex.md
LamarMoore: @martyngigg, In principle, we do not need to introduce a new widget type in the property widget factory, just careful handling of the property. Since the property is essentially composed of three properties, the GUI can be constructed using these properties independently, when the user confirms the selection in the dialog, the three properties will be set separately. The only stipulation being that a valid workspace is set at the point the user tries to extract indices from the property (because it needs to use a valid `IndexInfo`). See the prototype https://github.com/mantidproject/mantid/pull/19699

SimonHeybrock:
@LamarMoore I think the point is that the validation should happen in the validator such that users can get notified about, e.g., an out-of-range spectrum number. That is, the translation (and thus validation) of the indices should be done *before* requesting indices in `exec()`.

LamarMoore:
@SimonHeybrock @martyngigg correct me if I am wrong but isn't this taken care of by the `PropertyWithValue::isValid` method anyway?

SimonHeybrock:
@LamarMoore Well, you need to do (or attempt) the index translation in the validator, otherwise you will not know if the requested spectrum number 42 exists in the workspace.

LamarMoore:
@SimonHeybrock I do this in the `isValid()` but I do not store the result. I just call the `makeIndexSet` method and append more information to the error message. This mechanism could be better though. Check the `WorkspacePropertyWithIndex.tcc` file in the prototype.

SimonHeybrock:
@LamarMoore Where? I looked here https://github.com/mantidproject/mantid/pull/19699/files#diff-c62cf4bbd76ca5caab40ae1e565436a9R59 but that only seems to check each sub-property on its own?

SimonHeybrock:
Oh, I found it, is is encapsulated in the `getIndices()` method. 👍

PR #43, opened 2017-05-17
Tomography Python GUI Design

DTasev:
This is the high level design and implementation document for the Tomography GUI in Python.


JeanBilheux:
Because you are working with NeXus files, did you notice any file size issues? During the last ESS imaging workshop I heard a few people talking about the size of their imaging NeXus file and how big they were.

DTasev:
Working with `tiffs`, `fits` and `nxs` I haven't seen any file size difference for the same dataset.

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -0,0 +1,181 @@
...
+
+To answer the initial question, building the selection for the user is now just a matter of iterating the dict and getting the (user friendly) name, not the uuid, and displaying that.
+
+The unique ID will be generated using python's `uuid` package.
+
+A future upgrade on this will take into account which is the _current focus_ of the user, and when applying an operation will automatically select _that_ stack as the default choice.
+
+## Visualisation
+Visualisation will be done using Matplotlib. Each stack will be a separate part of a QDockWidget, so the users will be able to arrange the visualised stacks in any way they want. This also means that there can be more than one stack loaded at a time.
+
```
quantumsteve: Do you have a proof of concept verifying that Matplotlib is up to this task? Interactive viewing of tens of gigabytes will stress any any visualization package.

DTasev: image

This is drawn on a FigueCanvasQt (the drawing algorithms have not been changed), with a very crude 'slider' implementation, showing the last indice of a 24GB stack. Changing the indices is a bit slow (feels like 0.5s). The speed with which the image is shown does not seem to depend on the stack size (i.e. whether 100 or 15000 images), but on the image size -> images (512x512) are drawn a lot faster than (2048x2048). There might be a way to speed up the drawing.

RAM is also a big limit for large stacks, a solution for that is something that ImageJ provides called a Virtual Stack, and instead of loading every image in memory, it loads _only_ the currently requested slice. I have not yet done performance testing for the virtual stack

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -0,0 +1,182 @@
...
+<!-- /TOC -->
+
+# Implementation design for ISIS_Imaging GUI
+The user requirements for the interface have been listed [here](https://github.com/mantidproject/isis_imaging/wiki/High-Level-User-Requirements-and-Use-Cases) and have been reviewed and confirmed by the imaging scientists.
+
+The development requirements, following from the user requirements, are listed [here](https://github.com/mantidproject/isis_imaging/wiki/High-Level-Development-Requirements-and-Guidelines)
+
+# Motivation
+The current [Tomography Reconstruction](https://github.com/mantidproject/mantid/tree/043095a619bc8851942a31253a1a8f8b820ab30f/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography) interface was found to not satisfy the current imaging requirements. An easy solution to the problems was not found. With the agreement of the IMAT imaging scientists work is to be started on this design for a rewrite of the interface. Python is the language of choice, in order to be able to make use of visualising libraries like Matplotlib and reconstruction packages like Tomopy.
```
LamarMoore: With agreement from IMAT imaging scientists and senior members of the Mantid team.

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -0,0 +1,182 @@
...
+The Loading will be done via a dialogue:
+- It will ask for the paths for Sample, Flat and Dark images. Only Sample path is required and loaded. Dark and Flat images are NOT loaded, but the paths will be cached.
+- The user can specify indices in 3 spinboxes for [Start, Stop, Step]. The spinboxes cannot be negative. Boundaries are: 
+    - Max value of Stop is the number of files with the extension (detected dynamically after selection of the file). 
+        - Stop does not have a lower bound, meaning the user can select a Stop value lower than the Start value. Doing so will result in a ValueError. 
+        - This can be avoided with a signal triggered by changing the Start spinbox, that changes the setMinimum of the Stop spinbox. (low priority)
+    - Max value of Start is Stop - 1. 
+    - Min value of Step is 1, and max value of Step is the value of Stop.
+- The selection will require the user to select a file, NOT a directory.
+- The image format will be determined from the selected file.
```
LamarMoore: Will you include filtering based on a file type?

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -0,0 +1,182 @@
...
+To answer the initial question, building the selection for the user is now just a matter of iterating the dict and getting the (user friendly) name, not the uuid, and displaying that.
+
+The unique ID will be generated using python's `uuid` package.
+
+A future upgrade on this will take into account which is the _current focus_ of the user, and when applying an operation will automatically select _that_ stack as the default choice.
+
+## Visualisation
+Visualisation will be done using Matplotlib. Each stack will be a separate part of a QDockWidget, so the users will be able to arrange the visualised stacks in any way they want. This also means that there can be more than one stack loaded at a time.
+
+This must allow for the user to go to any indice from the volume.
```
LamarMoore: index -> indices

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -0,0 +1,182 @@
...
+This must allow for the user to go to any indice from the volume. 
+
+The visualisation must allow for a rectangle ROI selection, that is persistent if the image underneath is changed. It also must allow for the visualisation of a histogram of the selected ROI (computation should be done in the `core` package).
+
+Currently the idea is to use `FigureCanvasQtAgg` with the default drawing algorithms, for visusalisation of the images. Matplotlib provides a `RectangleSelector` class which will be used to do the ROI selection. If this proves to be too slow, a solution may be to tailor the drawing function, while still retaining the ROI selection functionality.
+
+## Applying a filter
+Applying a filter will bring up a dialogue in which the user has to select on which stack to apply the filter via a dropdown menu, and fill in the required parameters the filter has.
+
+Any filters should be dynamically registered using a [cli_registrator](https://github.com/mantidproject/isis_imaging/blob/master/isis_imaging/core/algorithms/cli_registrator.py) style approach.
```
LamarMoore: This link doesn't point to anything for me. Can you just expand on (add a brief explanation of) these things so readers don't have to keep switching documents?

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -0,0 +1,182 @@
...
+
+Current issues for this section:
+- Dynamic dialogue building
+  - Solution is to use an approach like the `cli_registrator`, but pass in the QObject in which the filters will register themselves in
+  - Dialogue registration is a bit more complicated, because we have to connect the dropdown menu to the dialogue's `.show()` method.
```
LamarMoore: Confused about what is meant to happen here. Do you have an image or a diagram?

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -0,0 +1,182 @@
...
+- Transferring information from the dialogue (the parameters) to the execution
+  - Current solution is to use a partial function, decorated with all of the parameters. No other function down the chain of execution will have to worry about the parameters that way.
+- Handling filters with multiple steps
+  - Background correction and contrast normalisation need to calculate values for scaling, do the filter execution, and then apply the scaling to the images.
+     - Currently the solution is to apply a monad style approach.
+
+## Undoing an operation
+Building the Undo operation may be complicated, but here are a few high level approaches:
```
LamarMoore: Could you store each image operation as a list and only apply to the image at the point of visualization? Then you could pop operations off the list (instead of managing crazy amounts of memory). Apologies if this is basically what you were trying to say here.

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -0,0 +1,182 @@
...
+            - But also provides the opportunity for the user to easily visualise different operations' results.
+
+## Histograms
+Computation of the histograms should be done only on the ROI, if no ROI is selected do the whole image. The computation itself should ideally take place in the `core.algorithms` package.
+
+Visualisation can be one of the two:
+- As a small widget inside the stack window
+- As a small independent widget floating around.
+
+A problem that needs to be considered is, should we allow for more than _one_ histogram to be active at a time? For example ImageJ does not.
```
LamarMoore: Is this a memory issue? e.g Is this because for a 16-bit image you have an array with 65536 entries? Otherwise you have to handle the histogram "window" being updated every time you select a new image so users know which image the histogram corresponds to.

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Processing a full volume
+A process list should provide the option to be applied to the whole stack locally. Details for the implementation need to be considerend alondside the [Undoing an operation](#undoing-an-operation) implementation. 
+
+The process list will store the operation(+params) that needs to be applied. Doing this for the whole stack would simply require walking through the process list. There is _nothing_ currently implemented in the `core` module that allows for that. 
+
+A new configuration might have to be added in `core.configurations` to handle this type of ordered filter application. A class that takes the name and dynamically imports/applies the operation might also be required.
+
+## Issues with Finding Center of Rotation and Reconstruction
```
LamarMoore: This section would really benefit with some images.

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -0,0 +1,182 @@
...
+
+Currently the automatic COR calculation is done on the sinograms and uses `core.imopr.cor`. Ideally we want to keep that behaviour. Slight change might be necessary to the `core.imopr.cor` module, as currently it does not return the CORs, but only prints them in the console.
+
+### Center of Rotation (COR) with imopr corwrite
+Once the sinogram issue has been resolved, we will have the sinograms in contiguous memory. This section will assume we already have solved that.
+
+The `core.imopr.corwrite` module works with sinograms. It saves out reconstructed slices with a range of CORs. We want to keep that behaviour as is, with the addition that after the process of saving out is finished, we visualise them back from the user, so they can select the best COR by seeing which is the best reconstructed slice.
+
+## Reconstruction
+Once the sinogram issue has been resolved, we will have the sinograms in contiguous memory. This section will assume we already have solved that.
```
LamarMoore: I think it is ok if you do not repeat this assumption. It is implied.

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -0,0 +1,182 @@
...
+
+The reconstruction works on sinograms and will be done via third party tools Tomopy, Astra Toolbox, MuhRec, etc.
+
+Specific dialogues might have to be created for each tool, to handle the different parameters each one has.
+
+## Remote submission
+The package needs to support submission to remote cluster. This section will be updated at a later point.
+
+### Remote compute resource used at ISIS: SCARF
```
LamarMoore: Don't you already have a prototype for this?

Design/ISIS_Imaging/ISIS_Imaging_general.md
Anders-Markvardsen: Also, I suggest rewording "Python is the language of choice, in order to be able to make use of visualising libraries like Matplotlib and reconstruction packages like Tomopy" to "Python is the preferred language here, in order to be able to make use of visualising libraries like Matplotlib and interactions with Python based reconstruction packages like Tomopy"

Anders-Markvardsen:
Please also add a section in your design doc how you will be adding automated test to this interface. Do you suggest it will follow the exact same automated testing design setup as the existing imaging interface

Anders-Markvardsen:
A graphical 'diff' between the old and new interface would be good, pointing out areas where the same look and feel is maintained between these interfaces and where this is not the case as broadly agreed with scientists. Last thought of today - would it be possible to reuse automated tests for the old GUI?

Design/ISIS_Imaging/ISIS_Imaging_general.md
DTasev: To come back to this question, Matplotlib with PyQt5 seems to do the visualisation about 5x faster for 2048x2048 images. It has been mentioned a possibility for 4096x4096 images in the future. 

I've scaled some images up to that size, and the native drawing algorithms seem to be performing well. It is not quite as fast as `ImageJ/FIJI`, but maybe there is a potential for optimisation.

I also scaled some images up to 8192x8192 and 16386x16386. Both matplotlib and ImageJ seem to struggle with that size, which I think is expected. We can still load them, but changing to another image takes ~1 second for 8k and ~5 seconds for 16k

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -1,61 +1,112 @@
...
 
 The development requirements, following from the user requirements, are listed [here](https://github.com/mantidproject/isis_imaging/wiki/High-Level-Development-Requirements-and-Guidelines)
 
-# Motivation
+# 1. Motivation
 
 The current [Tomography Reconstruction](https://github.com/mantidproject/mantid/tree/043095a619bc8851942a31253a1a8f8b820ab30f/MantidQt/CustomInterfaces/inc/MantidQtCustomInterfaces/Tomography) interface was found to not satisfy the current imaging requirements. An easy solution to the problems was not found. With the agreement of the IMAT imaging scientists and senior members of the Mantid team, work is to be started on this design for a rewrite of the interface. Python is the preferred language here, in order to be able to make use of visualising libraries like Matplotlib and interactions with Python based reconstruction packages like Tomopy.
 
-# User Requirements
+# 2. Old Tomography GUI Design
```
LamarMoore: It seems like this section combines what is missing in the old GUI and some elements of what you want to add to the new GUI. It may be better to separate the two.

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -109,72 +160,72 @@ Slices - referrs to a reconstructed sinogram. ![Reconstructed slice](https://git
...
 We can either work towards the package being installable using `python setup.py install`, making it available system wide. The alternative solution is to use it from the folder and provide global access like [the example here](https://mantidproject.github.io/isis_imaging/user_guide/setting_up.html#global-access-to-the-package)
 
-## Testing
+## 7.3. Testing
 
 Testing will use `nose` to run the tests and `coverage` to compute the tests' coverage. The actual unit tests should use the built-in `unittest`, and some might have to use `numpy.testing` for asserting equality for `numpy.ndarrays`.
 
 A module to run the tests has been provided and is called `run_tests.py`, in the root of the repository. It takes as an argument the names of different packages, and then runs the tests specifically for them. The possible arguments are listed via `-h` flag, or at the top of the file in the source.
 
 There is currently no integration with Mantid's testing.
```
LamarMoore: Is there a reason why it should be integrated with Mantid testing? Seems completely separate to me. Any other views?

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -109,72 +160,72 @@ Slices - referrs to a reconstructed sinogram. ![Reconstructed slice](https://git
...
 
-# Guidelines for ISIS Imaging CORE
+# 8. Guidelines for ISIS Imaging CORE
 
-## CUDA filter expansion
+## 8.1. CUDA filter expansion
 
 Integrating a package that supports CUDA has potential for improvement and expansion of the imaging filters.
 
 A potential problem for CUDA is for machines with low VRAM, transferring the data multiple times might be slower that just doing it in a bulk in the CPU, if it has a lot of RAM.
```
LamarMoore: Is there a way to query the amount of VRAM on a system and make a decision based on the result?

DTasev: Brief look at the CUDA API says yes -> [link to related topic](https://stackoverflow.com/questions/15966046/cudamemgetinfo-returns-same-amount-of-free-memory-on-both-devices-of-gtx-690)

There is already an algorithm in the `core` that can calculate the expected memory usage, we could make a decision based on that

Design/ISIS_Imaging/ISIS_Imaging_general.md
DTasev: Good spot, I have now moved the paragraphs to the related section https://github.com/mantidproject/documents/commit/8f228b37d9b145deab7742bfc2d0e14c185091c7

Design/ISIS_Imaging/ISIS_Imaging_general.md
```diff
@@ -0,0 +1,733 @@
...
+
+# 2. Old Tomography GUI Design
+
+- Note: In the images all things in red rectangles do not work. All things in blue rectangles DO work.
+
+The `Run` tab handles the selected tool and it's parameters, the logging into the compute resource for remote reconstruciton, and the directories for input images. All of these things can be provided through a File/Action menu to not clutter the screen, as all of them need to only be performed once.
+
+![Run Tab](https://github.com/mantidproject/documents/blob/tomography_gui/Design/ISIS_Imaging/runtab.png)
+
+![Filter Tab](https://github.com/mantidproject/documents/blob/tomography_gui/Design/ISIS_Imaging/filtertab.png)
```
DTasev: I've changed the image and added a more detailed write up :+1:

DTasev:
In an attempt to move the Python version requirement to 3.5, I have come across an extreme slowdown in initial memory allocation and processing performance. 

The reason has been outlined in http://bugs.python.org/issue30919, and it seems to be a behaviour change in `multiprocessing`'s shared memory allocation on Unix systems between python 2.7 and 3.5. 

- In 2.7 the memory was actually allocated in memory, and then shared with subprocesses. 
- In 3.5 the memory is written out as a file, which is then used as the 'shared memory' between subprocesses. The reason for that is that Python 3.x supports more ways of creating subprocesses, so this was the only way to support all of them.

However, as I do not think any of the additional ways of creating subprocesses are useful for this package, this change makes moving to 3.x impossible, until a workaround is found. I have taken the following steps as possible solutions:

- I've asked on the Python issue board if anyone has a way to force the python 2.7 behaviour in python 3.5, this would by far be the easiest solution, as no significant code change will be necessary. However it might require changes to the Python source code, which I am not happy about, as I want to use a standard python distribution
- I am looking into packages that can be used instead of the built-in `multiprocessing`, to handle the parallel processing. Some have already been added to the design document, like `dusk` and `numba`. The problem with this approach is that neither `dask` nor `numba` are available on SCARF right now, but they can be requested and added.

I will continue to look into this issue and will add a section to the design document to make sure this information is not lost

Anders-Markvardsen:
Lamar, Nick, Dimitar and myself discussed this design document, which also will serve as a handover document to Dan Nixon starting Sep. 

Dimitar will make some further updates before his leaves (18th Aug) but otherwise thanks and we are happy with it.

DTasev:
I've pushed changes to rename the package to MantidImaging, any changes are renaming and no content was added.

PR #46, opened 2017-06-08
make example const correct

quantumsteve:
One shouldn't be able to assign to a const object.

This also adds an example using C++17 structured bindings.


quantumsteve:
Closing because the code snippet no longer exists in master.

PR #48, opened 2017-07-12
Mantid as a SciPy styled Package Design

NickDraper:
Design document for us to harmonise our view on what we want to achieve under this strategic aim.

### DIFF
--- a/Design/Mantid_As_a _SciPy_Styled_Package.md
+++ b/Design/Mantid_As_a _SciPy_Styled_Package.md
@@ -1,0 +1,78 @@
+Motivation
+==========
+
+Shaping Mantid as a SciPy styled package is one of the strategic aims in the [ORNL Mantid 5 year plan](ORNL_Mantid_5yearplan.pdf).
+Howver it has become clear over several meetings and workshops that what should be included as part of this aim is not commonly agreed or 
+understood within the development team or management of the project.
+This document is an attempt to harmonise this understanding or at least provide a center point for discussion and agreement or what we want 
+to achieve as part of this strategic aim and to provide a baseline high level plan of what we need to do to implement it.
+
+Requirements
+============
+
+As mentoned above the requirements underlying the strategic aim are not universally shared so this document will serve as a place to detail
+what we have agreed.
+
+This is an initial draft from a discussion between Nick Draper and Martyn Gigg and will need to be discussed and agreed with 
+other team leads.
+
+Must
+----
+
+1. The Algorithms need to be seperated into sensible sub packages that allow sensible imports of only the areas of interest 
+   rather than all algorithms as is the current situation.
+1. The current top level api import must be maintained in name and function. 
+   So `from mantid.simpleapi import *` must import all of the sub packages.
+1. The documentation structure of Algroithms must match that of the sub packages to make it easier to locate the documents you need.
+1. The category structure in Mantid must match the python package structure, but use english language syntax rather than Python syntax.
+   For example Neutron vs Mantid.Neutron.
+   
+
+
+Should
+------
+
+1. The structure of the python sub packages and other structures should be based on that presented by Andre Savici at the 
+   [2017 Users Workshop] (../Presentations/DevMeetings/2017-06/Mantid4PythonLibrariesAndNames.md).  
+   Note: this will not include the changes to the naming of algorithms, just the package structuring.
+1. We should take advnatage of the restucturing into smaller packages to define a small number of these to be "core".
+   This will define this term, which currently is interpreted by each individual.
+1. The C++ library structure should mirror the python package structure as far as possible to make it easy for developers to 
+   navigate the code.
+1. We should decrease the time it takes to generate the Python API at run time. 
+   Investigations have shown this is a major part of the startup time of Mantid.
+   
+
+Could
+-----
+
+1. The developers will need tools or help to make it easier to add new libraries, and have full support (builds, python, installers, 
+   unit testing, docs, static analysis etc).
+1. While restructuring of the Algorithms libraries it would be beneficial to breakup the very large libraries of Algorithms and 
+   DataHandling.
+1. Low level functions for a limited list of operations (ConvertUnits, Rebin and possible a few others), could be created and
+   exposed to python allowing simple operations on vectors or numpy arrays of data without using workspaces or geometry etc.
+   1. Where these are created they should be used within the matching algorithm as long as that does not degrade 
+      performance or maintainability.
+   1. The low level function for ConvertUnits could replace much of the code in UnitConverter.py.
+
+Won't (recorded for the future)
+-------------------------------
+
+1. Renaming of algorithms to follow a more consistent pattern.  
+   This is not currently desirable due to the dirsuption it would make to users scripts.
+
+Design
+======
+
+
+
+
+Example
+-------
+
+Questions
+=========
+
+1. If you have imported a sub selection of packages of IO, what does that mean for Load?
+


peterfpeterson:
First big comment without thinking terribly long... If this is mantid as [scipy styled libraries](https://www.scipy.org/), the description misses out on data (NumPy for scipy; Workspaces for mantid) and user interface (ipython and matplotlib for scipy; widgets, custom guis, and workbench for mantid).

NickDraper:
I can add a bit about data structures, but on the grounds we are talking about libraries, I don't think we should discuss user interfaces at all.

I'm keen to keep this aspect clean from creeping into every other piece of work we do, as that way we will never be able to complete it, let alone agree.

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+### Algorithm and Fit Function Pages ###
+
+These will remain as is, except that the category links will be redirected to the module pages.
+
+C++ Library Structure
+---------------------
+
+In order to make the C++ library easy to navigate we will rearrange the library to match the module structure of the python modules.  This will increase the number of libraries we have, but will help split out some of those that have become bloated, such as Algorithms and DataHandling.
+
+?? Do we want separate C++ libaries for sub-modules such as mantid.math.axes or should be stop at a certain level and gather all the mantid.math into a single library? ??
```
martyngigg: As mentioned in a previous comment I think smaller libraries will allow us more flexibility and for instance would mean that we don't have to link many more dependencies than are required by that module.

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+
+If we decide to allow objects (such as algorithms) to be aliased in other python modules then we will need to define a rule of where the C++ algorithm will reside, we should use the first entry in the algorithm Category to be where the real code exists and the rest are aliases.
```
martyngigg: I think it is worth mentioning that we still wish to keep the rule that an algorithm name is unique globally to avoid accidental overwrites in Python.

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+Deciding which operations to expose like this should be constrained by the following guidelines:
+
+1. We should not replicate functionality that is directly available in NumPy, SciPy etc.
+1. All operations should work with inputs and outputs that are exported through simple python object and NumPy arrays.
+1. We should only consider heavily used core operations for exposing in advance of a specific user or developer need.
+
+Examples that we could expose:
+``` python 
+# For simple conversions of an array of values to another neutronic unit
+result_numpy_array = mantid.api.functions.convertUnits(input_numpy_array, target_unit, l1, l2, two_theta, energy_mode, e_fixed)
```
martyngigg: The only thing I don't like about this is the `.api` part. Why not have this in the same place as the `ConvertUnits` algorithm but maybe in a `lowlevel` or `functions` namespace?

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+# For rebinning a simple pair of matched arrays
+(data_numpy_array, boundaries_numpy_array) = mantid.api.functions.rebin(data_numpy_array, boundaries_numpy_array, rebin_params_list)
+```
```
martyngigg: Same comment as above.

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+
+With the convertUnits function we could rewrite TOF_Converter and remove a lot of the custom unit conversion code.
+
+Questions
+=========
+
+1. Mantid.IO should we split out NeXus, ISAW etc?
+1. Should mantid.simulations be mantid.neutron.simulations ?
+1. If you have imported a sub selection of packages of IO, what does that mean for Load?
+1. is it possible to extract the short module description from a docstring?
```
martyngigg: Yes, as commented above and I think we should do this.

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+# For rebinning a simple pair of matched arrays
+(data_numpy_array, boundaries_numpy_array) = mantid.api.functions.rebin(data_numpy_array, boundaries_numpy_array, rebin_params_list)
+```
+
+With the convertUnits function we could rewrite TOF_Converter and remove a lot of the custom unit conversion code.
+
+Questions
+=========
+
+1. Mantid.IO should we split out NeXus, ISAW etc?
```
martyngigg: Yes, as commented a few times above.

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+1. Should mantid.simulations be mantid.neutron.simulations ?
+1. If you have imported a sub selection of packages of IO, what does that mean for Load?
```
martyngigg: Good point. It depends how context sensitive we want `Load` to be. If they do `mantid.io.nexus.Load` then it would only load the nexus loading algorithms and `Load` would not be able to load `.raw` files for example but I don't think I would expect it to as it's been imported from the `nexus` package - this could get too complicated though...

I think this is down at the detail level but we need to think about what is the easiest for our users.

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+As mentioned above the requirements underlying the strategic aim are not universally shared so this document will serve as a place to detail
+what we have agreed.
+
+This is an initial draft from a discussion between Nick Draper and Martyn Gigg and will need to be discussed and agreed with 
+other team leads.
+
+Must
+----
+
+1. The Algorithms need to be separated into sensible sub packages that allow sensible imports of only the areas of interest
```
OwenArnold: To be clear, you're talking about python modules here for the ease of python import? To distinguish this from a code directory reorganisation (although it might involve that) or includes at the c++ level.

quantumsteve:
The primary goal of "Mantid as a SciPy styled library" is improve interoperability with the larger scientific python ecosystem. Someone who prefers manipulating data structures with numpy, using matplotlib for plotting, scipy and lmfit for fitting, Pandas for spreadsheet-like operations (and much more) should have a first-class interface.

How does this proposal get Mantid closer to that goal?

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+1. The Algorithms need to be separated into sensible sub packages that allow sensible imports of only the areas of interest 
+   rather than all algorithms as is the current situation.
+1. The current top level api import must be maintained in name and function. 
+   So `from mantid.simpleapi import *` must import all of the sub packages.
+1. The documentation structure of Algroithms must match that of the sub packages to make it easier to locate the documents you need.
+1. The category structure in Mantid must match the python package structure, but use english language syntax rather than Python syntax.
+   For example Neutron vs Mantid.Neutron.
+1. We must maintain backward compatability with the majority of users scripts, small changes are permissible if we are confident that this will not affect a significant number of users or scripts.  All breaking changes must be notified in advance of the release, ideally supporting deprecated aspects for at least one release.
+1. All documentation and training courses will need to be updated to match the new structures.
+
```
OwenArnold: Distribution is not mentioned here. Everything will still come in a single distribution including MantidPlot?

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+
+Should
+------
+
+1. The structure of the python sub packages and other structures should be based on that presented by Andre Savici at the 
+   [2017 Users Workshop](../Presentations/DevMeetings/2017-06/Mantid4PythonLibrariesAndNames.md).  
+   Note: this will not include the changes to the naming of algorithms, just the package structuring.
+1. We should take advantage of the restructuring into smaller packages to define a small number of these to be "core".
+   This will define this term, which currently is interpreted by each individual.
```
OwenArnold: I think "core" here is still undefined by this statement. Something like the following would be more useful: A package becomes "core" when ... {}

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+1. The C++ library structure should mirror the python package structure as far as possible to make it easy for developers to 
+   navigate the code.
+1. We should decrease the time it takes to generate the Python API at run time.
```
OwenArnold: This is a non-functional/operational requirement. Also, I would say that it more important that a "Should"

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+1. We should decrease the time it takes to generate the Python API at run time. 
+   Investigations have shown this is a major part of the start-up time of Mantid.
+   
+
+Could
+-----
+
+1. The developers will need tools or help to make it easier to add new libraries, and have full support (builds, python, installers, 
+   unit testing, docs, static analysis etc).
+1. While restructuring of the Algorithms libraries it would be beneficial to breakup the very large libraries of Algorithms and
```
OwenArnold: And to break the webs of this soft/runtime-dependency issue which makes the current breakdown of algorithm binaries meaningless.

peterfpeterson:
The feedback from Andrei's presentation was that it didn't go far enough. This proposal does not currently cover what is suggested by it's title.

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+    - `SubmitRemoteJob`, `AbortRemoteJob`
+
+**Facility Specific Libraries**
+
+* mantid.ess
+* mantid.hfir
+* mantid.ill
+* mantid.isis
+* mantid.sns
+* can add instrument or technique specific sublibraries
```
OwenArnold: I would suggest go further with this. In order not to to have highly facility specific things in say `mantid.neutrons.sans` I would suggest that Algorithms would live under `mantid.{facility}` until that have been proved to be useful/correct across sites. One common complaint that things in Mantid pretend to be general, when they are actually only implemented to work at one facility (possibly one instrument). Better to have more things in facility specific areas to reflect the reality.

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+    - `mantid.sns.corelli.CrossCorelation`
+
+### Generation ####
+
+The generation of the Mantid Python API for algorithms is currently done at run time on start-up, and has been found to take a large proportion of the time taken to start Mantid.  
+
+#### C++ Libraries of Algorithms ####
+
+Once C++ libraries are compiled within Mantid they will not change (other than being added to or removed from the plugins directory).  Therefore the API for a C++ library can be calculated once, and then reused until a new version that library is installed.
```
OwenArnold: I don't understand what this is trying to say. 

For non-plugins, `Kernel`, `API`, `Geometry`, `Beamline`,`Crystal`.... IMHO we should evaluate whether the current dynamic library approach is actually a good one.

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+This is an initial draft from a discussion between Nick Draper and Martyn Gigg and will need to be discussed and agreed with 
+other team leads.
+
+Must
+----
+
+1. The Algorithms need to be separated into sensible sub packages that allow sensible imports of only the areas of interest 
+   rather than all algorithms as is the current situation.
+1. The current top level api import must be maintained in name and function. 
+   So `from mantid.simpleapi import *` must import all of the sub packages.
```
quantumsteve: I wish `from mantid.simpleapi import *` returned something like `I'm sorry, Dave. I'm afraid I can't do that.`

That's probably a step too far, but we may want to require that an updated API nudges users towards Python best practices. The best we can do without breaking scripts is provide loud warnings, suggest code changes and continue running.

Design/Mantid_As_a _SciPy_Styled_Package.md
```diff
@@ -0,0 +1,259 @@
...
+1. The documentation structure of Algroithms must match that of the sub packages to make it easier to locate the documents you need.
```
LamarMoore: type `Algorithms`

PR #49, opened 2017-07-19
Workbench Design

martyngigg:


martyngigg:
Add a document on high-level thoughts on tasks done on `master` before work on the workbench takes off.

OwenArnold:
@martyngigg. Looks good. I have no further comments on this.

martyngigg:
@OwenArnold I've made a few slight changes to the directory layout to keep it a little more C++ centric. I think it would be better long term for the current C++ Qt stuff to remain independent on Python as it's one less dependency to worry about in those layers. The final `.dll` naming then follows a bit more of the ParaView style naming convention. Let me know what you think.

Design/Workbench/design-development.md
```diff
@@ -0,0 +1,22 @@
+# Development
+
+## Coding Style
+
+It is proposed we introduce a new Python style guide for Mantid. All new and refactored code should then follow this guide. The [IPython style guide][ipython-style] offers quite nice guidelines and it is suggested that we
+follow these guidelines where applicable, including a template file for new code that could be used in the `class_maker` to generate new files.
+
+## Static Analysis
+
+We will no longer be able to rely on a compiler to catch simply errors - it is therefore imperative that we use static analysis to its fullest extent. While `flake8` is good it is not as thorough as `pylint` and given our
```
NickDraper: Flake8 is also a lot faster, is the benefit of pylint worth the speed penalty?

Also I'm not too happy with the idea of Flake8 on some code and PyLint elsewhere (I could be reading this wrong though).

Design/Workbench/design-development.md
```diff
@@ -0,0 +1,22 @@
...
+target is maximum reliability I propose that, at least on the gui components, we reinstate `pylint` for each pull request. Its configuration should be updated to remove some warnings that are too pedantic. As we are
+starting from a clean slate with the workbench it should not be an issue to keep the warning level at 0. Special care should be taken by reviewers to assess when any warnings are suppressed to check if this is indeed valid.
+
+## Documentation
+
+User documentation for the workbench will be added to the standard `docs` build and organised into a navigible structure appropriately.
+
+**Question: Where does developer documentation for this belong?**
```
NickDraper: A category on the wiki?

Design/Workbench/design-layout.md
```diff
@@ -0,0 +1,149 @@
...
+------
+
+The framework package will remain separate and called `mantid`.
+
+The new ui package will be called `mantidui` and have the following submodules:
+
+ - mantidui.plotting: contain custom plotting code based on matplotlib, i.e keep/make current behaviour, custom toolbars, custom figure window
+ - mantidui.widgets: contain these set of reusable widgets used to build the workbench & its components
+
+The workbench will be called `mantidplot` and depend on `mantudui` & `mantid`.
```
NickDraper: Do we really want to call it MantidPlot?  That surely would cause confusion.  I'd go with MantidBench or Mantid Workbench for now.

Design/Workbench/design-layout.md
```diff
@@ -0,0 +1,149 @@
...
+=======
+
+This document describes the layout and high-level details of the new matplotlib-based workbench that will replace MantidPlot.
+
+Workflow
+========
+
+The new workbench will be developed on a long-running feature branch within the main [`mantid`][mantidrepo] repository. The exception to this is the new directory
+structure described below for the graphical components. These changes will be made on the `master` branch prior to any work starting on the workbench. This
+will minimize the effect of merge conflicts on existing components that are going to be reused.
```
NickDraper: Do we have any idea how much effort this will take to make the changes, update the cmake scripts and packaging and get master working again?

Design/Workbench/design-layout.md
```diff
@@ -0,0 +1,149 @@
...
+
+The following diagram indicates the proposed directory structure for the graphical components.
+
+```
+mantid.git
+   |-- Framework
+   |-- MantidPlot
+   |-- qt
+       |-- applications
+       |   |-- mantidworkbench
```
NickDraper: Do we want the 'main' ui hidden so far down the directory tree?

Design/Workbench/design-layout.md
```diff
@@ -0,0 +1,149 @@
...
+ - `/MantidQt/MantidWidgets/InstrumentView` --> `qt/widgets/instrumentview`
+ - `/MantidQt/RefDetectorViewer` --> `/qt/widgets/refdetectorview`
+ - `/MantidQt/SpectrumViewer` --> `/qt/widgets/spectrumviewer`
+ - `/MantidQt/SliceViewer` --> `/qt/widgets/sliceviewer`
+ - `/MantidQt/CustomDialogs` --> `/qt/widgets/plugins/algorithm_dialogs`
+ - `/MantidQt/DesignerPlugins` --> `/qt/widgets/plugins/designer`
+ - `/MantidQt/Python` --> `/qt/python`
+ - Remove Factory directory as it seems unecessary
+ - All resources, fonts, images etc, moved under qt/resources
+
```
NickDraper: is it worth stating where the ui files should be?

Design/Workbench/design-layout.md
NickDraper: Again, not sure about re-using the old name

NickDraper: Particularly as we want this to be side by side installable with the old ui for a while

Design/Workbench/design-arch.md
```diff
@@ -0,0 +1,241 @@
...
+            show()
+
+    mantidqt.InterfaceFactory.subscribe(HFIRPowderMetadata)
+```
+
+**Note: This obseletes the design in [this pull request](https://github.com/mantidproject/documents/pull/40).**
+
+# `mantidqt` library
+
+This library will form the guts if the new application and will contain most of the components that will be pieced together to form the workbench. The details of the built/installed layout are:
```
LamarMoore: "...the guts of..."

Design/Workbench/design-arch.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Figure Management
+
+It is desirable within the plotting framework to have a mechanism for controlling how figures are managed. For example, when `plot` is called without specifying a window then we must make a choice of either:
+
+ - create a new figure or
+ - merge/overplot with an existing figure.
+
+The initial propsosal is to borrow ideas from the [mslice][mslice] and have a controls on the figures so that users can indicate what is to be done with this window. In the figures above each figure contains a pair of
+buttons `Hold/Active` or `Held/Active` (depending on its state). These buttons control what happens when a plotting function is called without explicit intent as to which figure it should affect.
```
LamarMoore: I find Hold/Active less descriptive and more confusing than keep/make active.

martyngigg: @LamarMoore @NickDraper I feel this is something that, whatever we choose, it will be fine with some and not with others. I'm open to other suggestions if you have any.

I imagine initial user testing will give us some better feedback on this to react to.

Design/Workbench/design-arch.md
```diff
@@ -0,0 +1,241 @@
...
+
+- **Question: Where should the controls for script execution go? I would suggest just on the main menu, with appropriate shortcuts set.**
+
+
+# Automated Testing
+
+One of the most common complaints about the existing MantidPlot application is stability. We must take a proactive approach to automated testing within the new workbench. Where possible we will use common patterns such as
+[MVP][mvp-fowler] to facilitate automated testing of the GUI elements and provide a better seperation of concerns within the code.
+
+For existing code and code that would possibly be rendered more difficult to work with using MVP then the suggestion would be to supplement the pattern approach with testing via the [Qt Test][qttest] library. It allows
```
martyngigg: Good point. I'll add this.

Design/Workbench/design-arch.md
martyngigg: :+1:

Design/Workbench/design-arch.md
```diff
@@ -0,0 +1,241 @@
...
+  |   |-- mainwindow.py
+  |   |-- mainmenu.py
+  |   |-- splash.py
+  |-- config
+  |   |-- __init__.py
+  |   |-- defaults.py
+  |   |-- configdialog.py
+  |   |-- plotting.py
+  |   |-- firsttimesetup.py
+  |-- resources.qrc
```
martyngigg: @NickDraper I'm not sure I understand. I imagine this will also need its own `utils` package but the other concepts are not clear. Do you mean just add something to say that this is not final and can be extended as needed?

Design/Workbench/design-development.md
martyngigg: My main concern is we will now be developing a fully fledged application in Python and will not have a compiler to catch basic mistakes. `flake8` is okay but `pylint` does a far better job at genuine error detection.

Consider this:
```
def get_volumes():
    return [1, 2, 3, 4, 5]

def get_other_volumes():
    return [1, 2, 3, 4, 5, 6]

def do_something(volume):
    pass

def do_something_else(volume):
    pass

for volume in get_volumes():
    do_something(volume)

for volme in get_other_volumes():
    do_something_else(volume)
```

It contains a subtle bug that `pylint` detects out of the box yet `flake8` gives it a clean bill of health. Pylint's error:

```
W: 22,22: Using possibly undefined loop variable 'volume' (undefined-loop-variable)
```

The bug is that there is a typo in the the second loop variable so the loop actually uses `volume` from the previous loop definition as python 2 has terrible scope rules. A C++ compiler would not even compile this code.

One could argue that a unit test should pick up this type of thing but I think pylint's warning is much easier to interpret and find where the problem lies.

On the speed side of things I was thinking that we could do something similar to the doctests/system tests in that we run it combined with one of the builds so that we don't have to build separately - this is much of what caused the slowdown before.

@mantidproject/gatekeepers Any thoughts here?

Design/Workbench/design-layout.md
martyngigg: I would either leave it here or do away with the `applications` directory and have it there. To me anything different requires you to learn where it is anyway.

Design/Workbench/design-layout.md
martyngigg: Will do.

Design/Workbench/design-layout.md
peterfpeterson: Probably a day to do the work (by somebody that actually understands `cmake`) and a day for issues that are caused on build servers/developer systems.

Design/Workbench/design-layout.md
```diff
@@ -0,0 +1,149 @@
...
+       |   |   |-- ISISReflectometry
+       |   |       |-- ISIS_Reflectometry.py
+       |   |       |-- ISIS_Reflectometry
+       |   |       |   |-- ...
+       |-- python
+       |   |-- mantidqt # python mantidqt module
+       |       |-- reduction_gui # Python reduction gui framework
+       |       |-- widgets # would contain combine qtwidgetscommon.dll with sip exports to give mantidqt.widgets python library
+       |       |-- __init__.py
+       |-- widgets
```
peterfpeterson: Is the filepath limit of 260 characters on windows going to get us on this?

martyngigg: Good point.

I think a worse case scenario would be `master_incremental`. Its code gets clone to `C:\Jenkins\workspace\master_incremental\label\win7\mantid` so that gives us a ~50 character minimum stem. It looks like the `qt\python\mantidqt\reduction_gui` sub directory has the most nesting currently and taking one the files in the leaf directory would take us to ~125 characters at present. This would leave us with a headroom of ~135 characters. I *think* this should be okay even with the deepish directory structure that CMake creates in the build directory.

Design/Workbench/design-development.md
peterfpeterson: Run `pylint` against `master` and be quickly fix any found issues. `flake8` is more appropriate for pull requests because it gives consistent results.

Design/Workbench/design-development.md
peterfpeterson: http://developer.mantidproject.org/

Design/Workbench/design-arch.md
```diff
@@ -0,0 +1,241 @@
...
+### Sample Python Metadata File
+
+Below is an example of updating the current [HFIR powder GUI][hfir_startup_file] startup file to understand the new mechanism. It is written to be able to be started
+standalone from the command line or from within MantidPlot.
+
+```python
+import sys
+
+from HFIRPowderReduction import HfirPDReductionGUI
+from PyQt4 import QtGui
```
quantumsteve: PyQt5?

Design/Workbench/design-layout.md
martyngigg: I think this might be a bit optimistic. I would probably say at least 2-3 days to shuffle and get everything compiling. There's a fair bit of file shuffling and changes to things like include paths to deal with.

Design/Workbench/design-layout.md
```diff
@@ -0,0 +1,149 @@
...
+======================
+
+Python
+------
+
+The framework package will remain separate and called `mantid`.
+
+The new ui package will be called `mantidui` and have the following submodules:
+
+ - mantidui.plotting: contain custom plotting code based on matplotlib, i.e keep/make current behaviour, custom toolbars, custom figure window
```
martyngigg: That makes sense. I will add a list of these names for clarity.

Design/Workbench/design-arch.md
martyngigg: Thank you!

quantumsteve:
ParaView is switching to the "fusion" style on MacOS. It might be something we also want to consider for the new workbench.

http://public.kitware.com/pipermail/paraview-developers/2017-July/005486.html
http://blog.qt.io/blog/2012/10/30/cleaning-up-styles-in-qt5-and-adding-fusion/

martyngigg:
It does look cleaner. Certainly worth considering for Linux & OS X

Design/Workbench/design-development.md
ianbush: I would prefer a developer section in the rst docs as these are more likely to be updated with changes in the code than a wiki or separate repository. See for example http://docs.mantidproject.org/nightly/development/AlgorithmMPISupport.html.

ianbush:
It looks like the important points have been considered in detail, and the design looks great to me :+1: .

martyngigg:
Merging as it was accepted at the TSC meeting on [2017/08/01](https://github.com/mantidproject/documents/blob/master/Project-Management/TechnicalSteeringCommittee/meetings/2017/TSC-meeting-2017-08-01.md)

PR #53, opened 2017-09-19
Summary document for detector scan NeXus saving

ianbush:
For discussion and review.

### DIFF
--- a/Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
+++ b/Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
@@ -1,33 +1,39 @@
 Detector Scan NeXus Saving
 ==========================
 
+## Terminology
+
+In Mantid terminology detectors refer to the individual elements within a bank or tube, i.e. pixels. A bank or tube is referred to as a component. In this document Mantid detectors are referred to as 'detector(s)/pixel(s)'.
+
+In NeXus terminology a detector is an `NXdetector`, which might be a bank or a tube. Elements within are referred to as pixels. In this document NeXus detectors are referred to as 'NeXus detectors'.
+
 ## Background
 
-Workspaces in Mantid can now support detector scans, but these can not currently be saved in the Mantid NeXus Processed format. In related work the ESS are looking at saving a new style instrument definition in their raw files, to be read, and eventually saved by Mantid. The new style instrument should be able to save the scan information within itself.
+Workspaces in Mantid can now support detector/pixel scans, but these can not currently be saved in the Mantid NeXus Processed format. In related work the ESS are looking at saving a new style instrument definition in their raw files, to be read, and eventually saved by Mantid. The new style instrument should be able to save the scan information within itself.
 
 For now something similar could be stored separately in the Mantid NeXus processed files, that will be moved when the NeXus instrument definition is available. However, this would create legacy files for a short period, which would be best avoided.
 
 ## Technical Requirements
 
-### Detector Scan Types
+### Detector/Pixel Scan Types
 
-Two types of scan are now supported, normal scans where every detector can scan independently (asynchronously), and sync scans where every detector has the same set of scan intervals. Only synchronous scans are currently used within Mantid, for D20 and D2B at the ILL.
+Two types of scan are now supported, normal scans where every detector/pixel can scan independently (asynchronously), and sync scans where every detector/pixel or pixel has the same set of scan intervals. Only synchronous scans are currently used within Mantid, for D20 and D2B at the ILL.
 
 ### Information to be saved in Mantid Nexus Processed
 
 #### DetectorInfo
 
-* The position of each detector for each scan interval
+* The position of each detector/pixel for each scan interval
   * Stored in `DetectorInfo` as a 3-vector containing the absolute position
-* The rotation of each detector at each scan interval
+* The rotation of each detector/pixel at each scan interval
   * Stored in `DetectorInfo` as a quaternion containing the absolute rotation
-* The mask flag for each detector for each scan interval
-* For each detector position & rotation a mapping to the detector and scan interval that it refers to
+* The mask flag for each detector/pixel for each scan interval
+* For each detector/pixel position & rotation a mapping to the detector/pixel and scan interval that it refers to
   * The detector ID for each entry
   * The time index for each entry
 * Scan intervals - a set of start and end times need to be stored:
-  * Synchronous scan - a single set of scan intervals for every point in the detector scan
-  * Asynchronous scan - an independent set of scan intervals for every detector
+  * Synchronous scan - a single set of scan intervals for every point in the detector/pixel scan
+  * Asynchronous scan - an independent set of scan intervals for every detector/pixel
 
 #### SpectrumInfo
 
@@ -38,9 +44,9 @@ Two types of scan are now supported, normal scans where every detector can scan 
 
 ## Current Implementation
 
-Spectra to detector mapping is stored in `mantid_workspace_1/instrument/detector`. There are five fields, `spectra` which provides the spectrum numbers, `detector_count` which provides the number of detectors for each spectra, `detector_list` which provides a list of all the detectors, `detector_index`, which gives the entry point into `detector_list` for a given spectra and `detector_positions` which, for unknown reasons, gives the R, &theta; and &phi; coordinates of every detector.
+Spectra to detector/pixel mapping is stored in `mantid_workspace_1/instrument/detector`. There are five fields, `spectra` which provides the spectrum numbers, `detector_count` which provides the number of detectors/pixels for each spectra, `detector_list` which provides a list of all the detectors/pixels, `detector_index`, which gives the entry point into `detector_list` for a given spectra and `detector_positions` which, for unknown reasons, gives the R, &theta; and &phi; coordinates of every detector/pixel.
 
-The 'true' source for the detector positions and rotations is in `mantid_workspace_1/instrument/instrument_parameter_map/data`. They are only stored here if they differ from the base instrument.
+The 'true' source for the detector/pixel positions and rotations is in `mantid_workspace_1/instrument/instrument_parameter_map/data`. They are only stored here if they differ from the base instrument.
 
 ## Implementation Ideas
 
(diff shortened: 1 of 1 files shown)


Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
```diff
@@ -0,0 +1,129 @@
+Detector Scan NeXus Saving
+==========================
+
+## Background
+
+Workspaces in Mantid can now support detector scans, but these can not currently be saved in the Mantid NeXus Processed format. In related work the ESS are looking at saving a new style instrument definition in their raw files, to be read, and eventually saved by Mantid. The new style instrument should be able to save the scan information within itself, but for now something similar could be stored separately in the Mantid NeXus processed files, that will be moved when the NeXus instrument definition is available.
```
SimonHeybrock: This sounds like it would create the need to support legacy files, once we have Nexus instrument definition? Is there a way to avoid such a temporary solution?

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
```diff
@@ -0,0 +1,129 @@
...
+│   │   │   ├───scan_log
+│   │   │       ├───time (contains the time for each entry in value)
+│   │   │       ├───value (single number per time entry, describing the distance along 'vector')
+│   │   ├───position_scan_...
+│   ├───detector_2
+│   ├───...
+├───sample
+```
+
+An open question here, especially when considering conversion from the old IDFs, is how to define the component used for `detector_X`. Technically this could be anything from the whole instrument to individual pixels, but logically it should be something like a tube. The only requirement is that a detector has pixels with a fixed offset at all times.
```
SimonHeybrock: Can detectors commonly be described in such a regular way? How would this deal with calibration or mechanical tolerances? Would that require having scan information for all pixels individually?

What I am saying is that maybe we want a way to have per-bank scanning in Nexus but allow pixels within the bank to be non-equidistant? Does that make sense?

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
```diff
@@ -0,0 +1,129 @@
...
+To resolve the differences between the NeXus proposal and Mantid one option is to allow components to scan. This is similar to how `RectangularDetector` works. The changes required would be:
+
+* Add the scan interface in `DetectorInfo` to `ComponentInfo`
+  * E.g. in addition to `ComponentInfo::setPosition(detIndex)` we have `ComponentInfo::setPosition({detIndex, timeIndex})`
+  * A `ComponentInfo::setPosition()` call updates all of the child positions (same for rotations)
+* For setting positions, rotations or mask that can change with the time index, both `ComponentInfo` and `DetectorInfo` will require a check to see if the parent is scanning
+  * If the parent is scanning it should throw
+* The in memory representation is the same as before - the absolute position and rotation of all pixels are stored
+
+The performance impact of this is that every call to `DetectorInfo::position()` will have an extra branch, even for non-scanning workspaces.
```
SimonHeybrock: Would it? If we update positions when written you have extra branches in `setPosition()`, but wouldn't `position()` be unchanged?

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
```diff
@@ -0,0 +1,129 @@
...
+├───sample
+```
+
+An open question here, especially when considering conversion from the old IDFs, is how to define the component used for `detector_X`. Technically this could be anything from the whole instrument to individual pixels, but logically it should be something like a tube. The only requirement is that a detector has pixels with a fixed offset at all times.
+
+#### Differences Between NeXus and Mantid Approach
+
+In the `DetectorInfo` representation in Mantid each position, rotation and mask flag of each pixel at each time index is stored separately. The scans are done on a pixel-by-pixel basis, and the components can not currently do any scanning. Different pixels are allowed to be given different scan intervals even within a component.
+
+The NeXus instrument proposal would not support scanning at the pixel level. The offset entries would be the same for each time index, with the position entries giving the scan for a component.
```
OwenArnold: Can you clarify what you mean when you say the "offset entries would be the same for each time index"

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
```diff
@@ -0,0 +1,129 @@
...
+│   │   ├───x_pixel_offset (size is number of detectors)
+│   │   ├───y_pixel_offset (size is number of detectors)
+│   │   ├───z_pixel_offset (size is number of detectors)
+│   │   ├───...
+│   ├───detector_2
+│   ├───...
+├───sample
+```
+
+The suggested method for dealing with detector scans is to add entries as `NXTransformation`s to the detector entry. This is shown in the example hierarchy below. For each position and rotation axis a separate entry is required within the detector, such as `position_scan_x`, `position_scan_y`, `position_scan_z`, `rotation_scan_x`, `rotation_scan_y` and `rotation_scan_z`.
```
zjttoefs: That's NXtransformations (plural lower case t)

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
SimonHeybrock: I think it basically means that you cannot "scale" a bank via a scan, i.e., the pixel gap is constant in time.

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
```diff
@@ -0,0 +1,129 @@
...
+
+The performance impact of this is that every call to `DetectorInfo::position()` will have an extra branch, even for non-scanning workspaces.
+
+An alternative approach is to have something similar to `RectangularDetectors`. Here instead of storing the scan positions for each detector we would just store the position for the parent. Positions would need to computed on the fly to get the actual position for each detector. This would have the advantage of reducing memory use*, at the cost of more CPU work. This would be a step away from some of the Instrument 2.0 design principles.
+
+*Note that the memory overhead due to scanning is not the main issue for the ILL instruments, addressing the overhead from using histograms with a  single count (no ToF) is a higher priority.
+
+#### Outstanding Questions
+
+* If components scan, are detectors still allowed to scan independently?
```
OwenArnold: Some instrument definitions, see Polref for example, although true also of anything with monitors, has detectors that are not attached to any higher level components. That should be considered.

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
OwenArnold: Thanks, that makes sense.

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
zjttoefs: The above isn't 100% complete. I think what you want @SimonHeybrock is supported, but we should improve the example. @matthew-d-jones is in touch.

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
ianbush: Ideally yes, we should avoid any legacy files. The scan saving support is not urgent, ideally it would be good to have something on the months time scale, so it might be possible to avoid new legacy files.

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
ianbush: In the NeXus instrument there would be no requirement for pixels to be equidistant, offsets for each pixel from the centre are given in 3 dimensions. I'm not sure for rotations though.

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
ianbush: Yes, good point, that should be `setPosition()`.

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
ianbush: Yes, that should definitely be allowed, the monitor case is a clear example.

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
```diff
@@ -65,7 +66,7 @@ raw_data_1
 ├───sample
 ```
 
-The suggested method for dealing with detector scans is to add entries as `NXTransformation`s to the detector entry. This is shown in the example hierarchy below. For each position and rotation axis a separate entry is required within the detector, such as `position_scan_x`, `position_scan_y`, `position_scan_z`, `rotation_scan_x`, `rotation_scan_y` and `rotation_scan_z`.
+The suggested method for dealing with detector scans is to add entries as `NXtransformations` to the detector entry. This is shown in the example hierarchy below. For each position and rotation axis a separate entry is required within the detector, such as `position_scan_x`, `position_scan_y`, `position_scan_z`, `rotation_scan_x`, `rotation_scan_y` and `rotation_scan_z`.
```
OwenArnold: Really need to be careful with use of "detector" in document.

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
```diff
@@ -110,17 +111,20 @@ To resolve the differences between the NeXus proposal and Mantid one option is t
...
 
 An alternative approach is to have something similar to `RectangularDetectors`. Here instead of storing the scan positions for each detector we would just store the position for the parent. Positions would need to computed on the fly to get the actual position for each detector. This would have the advantage of reducing memory use*, at the cost of more CPU work. This would be a step away from some of the Instrument 2.0 design principles.
 
 *Note that the memory overhead due to scanning is not the main issue for the ILL instruments, addressing the overhead from using histograms with a  single count (no ToF) is a higher priority.
 
-#### Outstanding Questions
+#### Outstanding Question
 
-* If components scan, are detectors still allowed to scan independently? 
+* If components scan, are pixels within that component still allowed to scan independently?
```
OwenArnold: I don't think that it makes sense. A scan within a scan doesn't sound right to me.

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
```diff
@@ -110,17 +111,20 @@ To resolve the differences between the NeXus proposal and Mantid one option is t
...
+  * Possible use cases for this would be tubes that might change in time, for example temperature, during a scan. The calibration might then be time dependent.
```
OwenArnold: I can't speak speak for other facilities, but tube calibration isn't performed like this at ISIS to my knowledge. I don't like the idea of a scan within a scan.

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
```diff
@@ -0,0 +1,139 @@
...
+*Note that the memory overhead due to scanning is not the main issue for the ILL instruments, addressing the overhead from using histograms with a  single count (no ToF) is a higher priority.
+
+#### Outstanding Question
+
+* Should pixels be allowed to scan, given that the parenting is not scanning?
+  * Possible use cases for this would be tubes that might change in time, for example temperature, during a scan. The calibration might then be time dependent.
+  * If the answer to this is that we do need per-pixel scans then how can we support this in the NeXus file? A different proposal may be required.
+  * Do we then need a mechanism in Mantid to define which components can scan, and which can not?
+
+This is the key question to answer - if we restrict the scans in this way we should be able to use the NeXus Instrument proposal as it stands. If we do not impose this restriction we will likely need something to allow the values in `x_pixel_offset` and similar to be set per time interval.
```
matthew-d-jones: `NXlog` can be used anywhere that time dependence is required for a dataset. So it would certainly be possible to use `NXlog`s for the pixel offsets if required.

ianbush: Would this require an `NXlog` for each pixel offset, or can you give an array of values?

matthew-d-jones: An array is fine, from the documentation: "Array of logged value, such as temperature. If this is a single value the dimensionality is nEntries. However, NXlog can also be used to store multi dimensional time stamped data such as images. In this example the dimensionality of values would be value[nEntries,xdim,ydim]."

ianbush: This is good news then - this fills in the gap I thought was missing. I'll make some updates to the doc to mention this.

Design/Instrument-2.0/Detector_Scan_NeXus_Saving.md
```diff
@@ -92,13 +92,31 @@ raw_data_1
...
+│   │   ├───x_pixel_offset (`NXlog` entry)
+│   │       ├───time (contains the time for each entry in value)
+│   │       ├───value (2D array, x offset for each pixel at each scan point)
+│   │   ├───y_pixel_offset (offset for each pixel)
+│   │   ├───z_pixel_offset (offset for each pixel)
+│   │   ├───...
+│   ├───detector_2
+│   ├───...
+├───sample
+```
```
OwenArnold: 👍

ianbush:
I am merging this as there have been no further updates recently.

PR #54, opened 2017-09-21
First draft of notes on issues with current workspaces.

SimonHeybrock:
As discussed during last core-team call this is a starting point for discussion on the Mantid workspaces.

### DIFF
--- a/Project-Management/CoreTeam/workspace-notes.md
+++ b/Project-Management/CoreTeam/workspace-notes.md
@@ -17,6 +17,7 @@ Focussing on `MatrixWorkspace`. At present numbering below is not an indication 
      - Not much that a base Workspace actually gives you, and it has lead to the widespread usage of down-casting.
    - Base class `MatrixWorkspace` for `Workspace2D` and `EventWorkspace` was meant to unify handling of the latter two workspaces in algorithms. In reality many algorithms simply have two big blocks for handling them individually.
    - The base class of `MatrixWorkspace`, `ExperimentInfo` is a complicated and holds a lot of complexity related to setting the instrument and loading parameter maps.
+   - A `MatrixWorkspace` is also an `IMDWorkspace` so has a dimension api as well as axes. Confusing. Should not have made a `MatrixWorkspace` and `IMDWorkspace`.
      - Provides a detector grouping mechanism based on externally set grouping (i.e., *not* the grouping used by `MatrixWorkspace`), used by `MDWorkspace`.
    - Performance as highlighted by Steve [here](https://github.com/mantidproject/documents/files/1383875/2017-03-18-Highlights.pdf)
 1. False concept that a Workspace subtypes are a pluggable thing. They really aren't and that has lead to `MatrixWorkspace` having to support very possible usage you might want for a data structure
--- a/Project-Management/CoreTeam/workspace-notes.md
+++ b/Project-Management/CoreTeam/workspace-notes.md
@@ -11,7 +11,9 @@ Focussing mainly on `MatrixWorkspace`.
    - Detector IDs can be set arbitrarily on `ISpectrum`, no check if this is in the instrument.
 1. `MatrixWorkspace` is meant to store histograms, but is used for many other purposes:
    - Single values.
-   - Single data points in each histogram. Leads to performance problems due to massive overhead from using histograms for this purpose, e.g., at ILL.
+   - Single data points in each histogram.
+     - Leads to performance problems due to massive overhead from using histograms for this purpose, e.g., at ILL.
+     - Need to use `Transpose` algorithm too often for processing/plotting, cutting connection to instrument.
    - Masking.
    - Grouping.
    - HKL values.
@@ -39,5 +41,14 @@ Focussing mainly on `MatrixWorkspace`.
 1. Bloated interface with an abundance of methods for different purposes, but lacking other essentials.
    - Different set of methods in each base class and derived classes.
    - No iterator support.
+     - Should be usable with something like `std::transform` together with a good set of mid-level operations.
+   - No proper definition of equality.
 1. Composite pattern provided by `WorkspaceGroup` is maybe too cumbersome and not flexible enough.
+   - Horribly tied into the Analysis Data Service, i.e., cannot be used without.
+   - Wasteful for multi-period workrpsaces since mcu hof the information is duplicated.
 1. Provides few invariants that client code can rely on.
+1. No way to pickle/serialize a workspace in python.
+   - Increasingly required for experiment control / data reduction interop.
+1. No way of defining a region of interest and inconsistent handling of masking.
+   - Masking or bin masking may or may not be respected by algorithms.
+   - Generalized 'selection' object attachable to workspace?
--- a/Project-Management/CoreTeam/workspace-notes.md
+++ b/Project-Management/CoreTeam/workspace-notes.md
@@ -2,13 +2,24 @@
 
 ## Issues with the current workspace
 
-Focussing mainly on `MatrixWorkspace`.
+Focussing mainly on `MatrixWorkspace`. At present numbering below is not an indication of seriousness or priority. 
 
 1. Constructed iteratively, i.e., workspaces start their life in an incomplete and dysfunctional state, vital information is added only later (size, instrument, spectrum to detector mapping).
 1. No encapsulation for data it contains, thus violating invariants:
    - Access to X, Y, and E makes breaking or abusing internal data easy.
    - Access to spectrum numbers on `ISpectrum` prevents validation (e.g., duplicates).
    - Detector IDs can be set arbitrarily on `ISpectrum`, no check if this is in the instrument.
+1. No support for mid-level concepts. Too few building blocks. i.e. `std::transform` on a workspace. Probably explains why Algorithms have become the basic building block in Mantid rather than being seen as they should be - the thing that allows you to expose your functionality to users via python and the workbench.
+1. Not enough workspaces types (thus the aforementioned misappropriation of `MatrixWorkspace`).
+   - In the current way workspaces and algorithms are built, adding new workspaces types is not very useful since existing algorithms will not work for them.
+1. The virtual inheritance issues
+   - Often `dynamic_cast` has to be used when working with workspaces, making client code bloated.
+     - Not much that a base Workspace actually gives you, and it has lead to the widespread usage of down-casting.
+   - Base class `MatrixWorkspace` for `Workspace2D` and `EventWorkspace` was meant to unify handling of the latter two workspaces in algorithms. In reality many algorithms simply have two big blocks for handling them individually.
+   - The base class of `MatrixWorkspace`, `ExperimentInfo` is a complicated and holds a lot of complexity related to setting the instrument and loading parameter maps.
+     - Provides a detector grouping mechanism based on externally set grouping (i.e., *not* the grouping used by `MatrixWorkspace`), used by `MDWorkspace`.
+   - Performance as highlighted by Steve [here](https://github.com/mantidproject/documents/files/1383875/2017-03-18-Highlights.pdf)
+1. False concept that a Workspace subtypes are a pluggable thing. They really aren't and that has lead to `MatrixWorkspace` having to support very possible usage you might want for a data structure
 1. `MatrixWorkspace` is meant to store histograms, but is used for many other purposes:
    - Single values.
    - Single data points in each histogram.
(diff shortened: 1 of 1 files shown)


soininen:
I have another issue: It is relatively easy to process the data histogram-wise, but in the 'vertical direction', things become complicated. I, for one, need to use the `Transpose` algorithm too often which cuts the connection to the instrument. Also, things like plotting data column-wise or accessing the vertical axis are painful to deal with.

Project-Management/CoreTeam/workspace-notes.md
```diff
@@ -0,0 +1,43 @@
...
+1. Often `dynamic_cast` has to be used when working with workspaces, making client code bloated.
+1. Pulls in around 120 include files, thus contributing a significant part to compile times for around 300 algorithms (plus their unit tests) that depend on it.
+   - Compile times have been highlighted repeatedly as an issue by the development team.
+1. Is made for holding "spectra", but is used for transformed data like in `SofQW`.
+   - Data is still histograms, but not mappable to detectors, spectrum-detector mapping is useless.
+   - Actual meaning of spectra is not supported by workspace, i.e., histogram workspace not in detector space, so spectra should not be mapped to positions via `SpectrumInfo` but to Q.
+   - In general, there may be other workspaces containing histograms that do not map to detectors of an instrument.
+1. Data in base classes, extended in child class.
+   - Data both in base classes and child classes makes design brittle.
+   - Should favor composition over inheritance?
```
OwenArnold: Yes definitely.

Project-Management/CoreTeam/workspace-notes.md
```diff
@@ -0,0 +1,43 @@
...
+1. Bloated interface with an abundance of methods for different purposes, but lacking other essentials.
+   - Different set of methods in each base class and derived classes.
+   - No iterator support.
```
OwenArnold: Also. No proper definitions for equality. I feel that this is a missing piece of many of the concepts in Mantid. No easy way to ask if `a` is equal to `b` without going through an Algorithm. I do appreciate that for some operations we need to match with a tolerance, but I would still like to see `==` operators overloaded more widely.

Project-Management/CoreTeam/workspace-notes.md
```diff
@@ -0,0 +1,43 @@
...
+1. Composite pattern provided by `WorkspaceGroup` is maybe too cumbersome and not flexible enough.
```
OwenArnold: Yes. And still horribly tied into the Analysis Data Service. Also `WorkspaceGroup` is wasteful for multi-period workspaces where much of the information is duplicated. This gives some indication of the workspacegroup/ads interdependencies that need to be considered https://github.com/mantidproject/documents/blob/19cfa130144a73239658dff4d532fcfc147ba3de/Testing/MantidPlot%20Workspace%20Dock%20Test%20Scenarios.md fortunately @LamarMoore has already codified this the characterise the existing behaviour.

Project-Management/CoreTeam/workspace-notes.md
```diff
@@ -0,0 +1,43 @@
...
+   - Single values.
+   - Single data points in each histogram. Leads to performance problems due to massive overhead from using histograms for this purpose, e.g., at ILL.
+   - Masking.
+   - Grouping.
+   - HKL values.
+1. Not enough workspaces types (thus the aforementioned misappropriation of `MatrixWorkspace`).
+   - In the current way workspaces and algorithms are built, adding new workspaces types is not very useful since existing algorithms will not work for them.
+1. The base class of `MatrixWorkspace`, `ExperimentInfo` is a complicated and holds a lot of complexity related to setting the instrument and loading parameter maps.
+   - Provides a detector grouping mechanism based on externally set grouping (i.e., *not* the grouping used by `MatrixWorkspace`), used by `MDWorkspace`.
+1. Very complicated to copy / create output workspaces.
```
OwenArnold: No way to pickle/serialize a workspace in python. This is increasingly required for ECP/Data Reduction interop.

Project-Management/CoreTeam/workspace-notes.md
```diff
@@ -0,0 +1,43 @@
+# Workspace
+
+## Issues with the current workspace
+
+Focussing mainly on `MatrixWorkspace`.
+
+1. Constructed iteratively, i.e., workspaces start their life in an incomplete and dysfunctional state, vital information is added only later (size, instrument, spectrum to detector mapping).
```
OwenArnold: I think your point 1. Must be fixed in any new design. This default-construct & later initialise pattern used in Mantid is not good.

Project-Management/CoreTeam/workspace-notes.md
```diff
@@ -0,0 +1,43 @@
...
+   - Detector IDs can be set arbitrarily on `ISpectrum`, no check if this is in the instrument.
+1. `MatrixWorkspace` is meant to store histograms, but is used for many other purposes:
+   - Single values.
+   - Single data points in each histogram. Leads to performance problems due to massive overhead from using histograms for this purpose, e.g., at ILL.
+   - Masking.
+   - Grouping.
+   - HKL values.
+1. Not enough workspaces types (thus the aforementioned misappropriation of `MatrixWorkspace`).
+   - In the current way workspaces and algorithms are built, adding new workspaces types is not very useful since existing algorithms will not work for them.
+1. The base class of `MatrixWorkspace`, `ExperimentInfo` is a complicated and holds a lot of complexity related to setting the instrument and loading parameter maps.
```
OwenArnold: I think the whole virtual inheritance approach should be reconsidered. There is not much that a base `Workspace` actually gives you, and it has lead to the widespread usage of down-casting (which you have already mentioned). Furthermore, I think we should drop the concept that a `Workspace` subtypes are a pluggable thing. They really aren't (at least in their current form) and that has lead to a lot of shoe-horning, maybe explaining the bloated interfaces you have already mentioned.

OwenArnold:
@martyngigg I think at one of the Mantid Dev Workshops you started looking at a Workspace redesign. I think what @SimonHeybrock has done here is useful, considering the existing problems first before looking at the solutions. This is just a placeholder to remember to link in what has already been considered as part of that workshop after the "requirements" have been gathered.

Project-Management/CoreTeam/workspace-notes.md
```diff
@@ -0,0 +1,43 @@
...
+
+1. Constructed iteratively, i.e., workspaces start their life in an incomplete and dysfunctional state, vital information is added only later (size, instrument, spectrum to detector mapping).
+1. No encapsulation for data it contains, thus violating invariants:
+   - Access to X, Y, and E makes breaking or abusing internal data easy.
+   - Access to spectrum numbers on `ISpectrum` prevents validation (e.g., duplicates).
+   - Detector IDs can be set arbitrarily on `ISpectrum`, no check if this is in the instrument.
+1. `MatrixWorkspace` is meant to store histograms, but is used for many other purposes:
+   - Single values.
+   - Single data points in each histogram. Leads to performance problems due to massive overhead from using histograms for this purpose, e.g., at ILL.
+   - Masking.
```
soininen: Defining a region-of-interest is another sore point in Mantid. Every algorithm has their own way of asking for and handling histogram and bin ranges, and you never know if your masking is respected or not. To be frank, the whole masking thing seems to be just an afterthought. Maybe something like a generalized 'selection' object which you could attach to a workspace to indicate which parts of the data, e.g. which histograms, which bins/X ranges should be touched by the algorithms would help here.

Project-Management/CoreTeam/workspace-notes.md
```diff
@@ -0,0 +1,43 @@
+# Workspace
+
+## Issues with the current workspace
+
+Focussing mainly on `MatrixWorkspace`.
+
+1. Constructed iteratively, i.e., workspaces start their life in an incomplete and dysfunctional state, vital information is added only later (size, instrument, spectrum to detector mapping).
+1. No encapsulation for data it contains, thus violating invariants:
+   - Access to X, Y, and E makes breaking or abusing internal data easy.
```
soininen: Sometimes I feel like we're missing basic mid-level functionality. There are the algorithms which usually do very high level things, but internally they get their hands dirty with `std::vector`s sometimes forgetting error propagation or masking while dealing with partial bins each their own way. Having a good set of standard arithmetic operations between the mid-level data types (histograms) would help a lot here.

martyngigg:
@OwenArnold Yes we did have a go at something. The notes are at https://www.mantidproject.org/DevWorkshop2016_Thu_small_group#Workspaces

Project-Management/CoreTeam/workspace-notes.md
SimonHeybrock: "Every algorithm has their own way of asking for and handling histogram and bin ranges" -> see work by Lamar done for ESS: http://docs.mantidproject.org/nightly/concepts/IndexProperty.html. We now *do* have such a unified way, but it would need to be rolled out to algorithms.

Project-Management/CoreTeam/workspace-notes.md
SimonHeybrock: Yes, basically you would want to run something like `std::transform` on an algorithm. I had once started implementing this, see https://github.com/mantidproject/mantid/pull/15465, but have not developed the approach any further since, so I would not recommend using it more at this point.

SimonHeybrock:
I have updated the notes based on comments I got so far. @OwenArnold I did not split this into sections since a lot is quite generic for all workspaces types, and, more importantly, we actually may want to get away from the current classification.

Project-Management/CoreTeam/workspace-notes.md
martyngigg: True. How does this relate to passing workspaces between MPI ranks too?

Project-Management/CoreTeam/workspace-notes.md
martyngigg: Agreed. I think our initial try at the developer workshop was good first attempt at this.

Project-Management/CoreTeam/workspace-notes.md
```diff
@@ -0,0 +1,54 @@
...
+   - In general, there may be other workspaces containing histograms that do not map to detectors of an instrument.
+1. Data in base classes, extended in child class.
+   - Data both in base classes and child classes makes design brittle.
+   - Should favor composition over inheritance?
+1. Bloated interface with an abundance of methods for different purposes, but lacking other essentials.
+   - Different set of methods in each base class and derived classes.
+   - No iterator support.
+     - Should be usable with something like `std::transform` together with a good set of mid-level operations.
+   - No proper definition of equality.
+1. Composite pattern provided by `WorkspaceGroup` is maybe too cumbersome and not flexible enough.
```
martyngigg: Agreed. A better way of "linking" datasets in general would be desirable so that you don't have to depend on any kind of string name.

Users are quite used to be able to collect together related workspaces and doing things on them as a whole so that type of thing would need to survive in some form or another.

Project-Management/CoreTeam/workspace-notes.md
SimonHeybrock: This is still unclear but the way I see things currently we will usually *not* pass complete workspaces between MPI ranks but just, e.g., a vector of histograms. All other parts of the workspace are available on all MPI ranks.

Project-Management/CoreTeam/workspace-notes.md
quantumsteve: Some users will consider this a feature instead of a bug. Sometimes we just need to say it's your foot.

Project-Management/CoreTeam/workspace-notes.md
```diff
@@ -0,0 +1,55 @@
...
+   - Horribly tied into the Analysis Data Service, i.e., cannot be used without.
+   - Wasteful for multi-period workspaces since most of the information is duplicated. 
+1. Provides few invariants that client code can rely on.
+1. No way to pickle/serialize a workspace in python.
```
quantumsteve: We should separate the ability to serialize a workspace from how function call in Python.

Project-Management/CoreTeam/workspace-notes.md
quantumsteve: Performance is another reason to be careful with virtual inheritance. Changes to vtkDataSet to avoid virtual functions in tight loops improved performance by a factor of 2-3x.

[2017-03-18-Highlights.pdf](https://github.com/mantidproject/documents/files/1383875/2017-03-18-Highlights.pdf)

Project-Management/CoreTeam/workspace-notes.md
martyngigg: I agree with @quantumsteve, we cannot take away the ability of users to get access and manipulate the raw data - it limits their scope for playing with reduction/analysis.

SimonHeybrock: I think this is not meant to take away that functionality completely. For example, if we let user modify `Y` directly, they might change its size without also changing the length of `X` and `E` (this was possible before adding the `Histogram` related types which now prevent this, but think of similar cases).
Instead we should simply allow access to the histogram, but not the internals of histogram *on the workspace interface*. If users really want to modify `Y` directly they would need to do that via the interface of the histogram and then use for example `ws.setHistogram(mydata)`.

AndreiSavici:
You might want to add a comment about dimensions: it contains a similar interface to MD workspaces in addition to own axes.
The other thing that I find hard is manipulate events in event workspaces from Python (i.e. I want to add a weighted event)

PR #56, opened 2018-05-24
Design document for workbench plot selector

ianbush:
Design document for the plot selector idea in the new workbench. PR is to allow discussion on the document.


Design/Workbench/Plot-Selector/design-plot-selector.md
```diff
@@ -0,0 +1,66 @@
...
+
+For consistency the plot selector widget should follow a similar design to the workspace selector. Double-clicking a plot name will bring the plot to the foreground and make it the active one.
+
+![Plot Selector Widget](plot_selector.png)
+
+**All plots created from the workbench are shown in the list.** Double-clicking a plot will bring it to the foreground. Filtering works as for the workspace selector. Close can act on one or more selected plots.
+
+![Menu Options](menu_options.png)
+
+**This figure shows the menu options for New, Arrange, Sort and Export.** New allows a choice of plot types, the same selection as for right-clicking a workspace. Arrange allows all plot windows to be cascaded or tiled. Sort allows sorting by name or last modified, ascending or descending. Export allows all selected plots to be exported to commonly used file types.
```
martyngigg: I like the idea of this matching how the workspace list works. By name here would you mean the title of the plot?

The other question is how we would deal with the `last modified` - does matplotlib store this type of information?

Design/Workbench/Plot-Selector/design-plot-selector.md
```diff
@@ -0,0 +1,66 @@
...
+
+Nice to have:
+
+* Selecting one or more plots and clicking close closes selected plots
+* Right-click menu option to close plots
+* Right-click menu option to rename plot
+* Option to sort by either name or last modified plot, ascending or descending
+* Export to export plots a useful subset of supported types
+* Right-click menu to export plot
+* New option to create a plot from the plot selector
```
martyngigg: Maybe use the delete key as well for closing plots like the workspace list viewer?

Design/Workbench/Plot-Selector/design-plot-selector.md
```diff
@@ -0,0 +1,66 @@
+# Plot Selector Widget
+
+## Requirements
+
+The request from the January 2018 user meeting was to have a widget that allowed easy switching between plots, in particular for SANS and Reflectometry users. Another request from the ILL was to have a method to easily arrange the plots.
+
+For consistency the design should follow as closely as possible similar functionality in Mantid, such as the workspace selector and plotting options within the workspace selector.
```
martyngigg: Good call on following a similar design to the workspace selector.

Design/Workbench/Plot-Selector/design-plot-selector.md
```diff
@@ -0,0 +1,66 @@
...
+Nice to have:
+
+* Selecting one or more plots and clicking close closes selected plots
+* Right-click menu option to close plots
+* Right-click menu option to rename plot
+* Option to sort by either name or last modified plot, ascending or descending
+* Export to export plots a useful subset of supported types
+* Right-click menu to export plot
+* New option to create a plot from the plot selector
+* Arrange option to tile or cascade the current open plots (this could be investigated further to see if anything nicer can be done here to arrange the plots)
```
martyngigg: Another option to add to the list would probably be to create a `matplotlib` `subplot` in a single window of a selected group of plots.

Design/Workbench/Plot-Selector/design-plot-selector.md
ianbush: Yes, I think the name would be the title of the plot/figure window. Currently these are Figure 1, Figure 2 etc. but I guess we should follow something like the MantidPlot convention to give these a sensible automatic name.

As far as I know matplotlib does not store this information, though we can see if is in there. It might be tricky to keep track of, so this needs investigation to see if it is possible.

Design/Workbench/Plot-Selector/design-plot-selector.md
ianbush: :+1:

Design/Workbench/Plot-Selector/design-plot-selector.md
ianbush: I like this idea, then everything is mirroring the workspace selector very closely.

Design/Workbench/Plot-Selector/design-plot-selector.md
NickDraper: I would be tempted to move the selection of whether to plot error bars to the following dialog when you are selecting the workspace to plot.

We don't do this with the right click menu for workspaces, but that is because we do not intend on showing any more dialogs.

Something similar to the plot advanced dialog.
image

NickDraper: It might be worth considering something like this
image


 where the right hand side is the Advanced plot gui that we have to build anyhow, the only addition would be the workspace list (and filter for when they have lots of workspaces).

ianbush: Yes, this seems like a good idea. Instead of `New` being a menu it can just be a button to bring up such an advanced plot dialogue.

martyngigg: That feels good but possibly becoming an extension now rather than part of the original scope. I think the the list, filtering etc is valuable without requiring the new dialog. If you have time then maybe that's something you could also look at after this?

ianbush: Yes, I agree with this. I think the new dialogue can be pushed into the 'nice to have' list. In any case this should be implemented after the advanced plot window is available, as this would be very much a secondary way of accessing it.

Design/Workbench/Plot-Selector/design-plot-selector.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Functional Design
+
+### Interface Design
+
+For consistency the plot selector widget should follow a similar design to the workspace selector. Double-clicking a plot name will bring the plot to the foreground and make it the active one.
+
+![Plot Selector Widget](plot_selector.png)
+
+**All plots created from the workbench are shown in the list.** The name shown will be the figure window title. Double-clicking a plot will bring it to the foreground. Filtering works as for the workspace selector. Close can act on one or more selected plots. Additionally plots can be closed individually by clicking the close button on one of the plots in the list.
```
AndreiSavici: Do we create unique names? If yes, is it easy for the user to select the correct plot just from the name? For example, I can have plot1 containing workspace w, spectrum 1, plot 2 can have color fill plot of the same workspace, and plot 3 also spectra 1 of the same workspace. If the figure names are `w-1`, `w-2`, `w-3`, selecting plots by name would be kind of difficult.

Design/Workbench/Plot-Selector/design-plot-selector.md
```diff
@@ -0,0 +1,73 @@
...
+
+![Menu Options](menu_options.png)
+
+**This figure shows the menu options for Group, Arrange, Sort and Export.** Group allows figures to be grouped, either as a set of subplots in a single figure or over-plotted in a single figure. Arrange allows all plot windows to be cascaded or tiled. Sort allows sorting by name or last modified, ascending or descending. Export allows all selected plots to be exported to commonly used file types.
+
+![New Plot Selection](new_plot.png)
+
+**Selecting a new plot will then prompt for which workspace to plot, and offer advanced plotting options.** Selecting multiple workspaces should give the same behaviour as selecting multiple workspaces in the workspace selection widget (e.g. same plot for 1D, side-by-side for 2D). The advancded plot options should follow those in the current MantidPlot.
+
```
AndreiSavici: :+1:

Design/Workbench/Plot-Selector/design-plot-selector.md
ianbush: A good question, I have been assuming we do have unique names. In MantidPlot the current behaviour is to do `[ws_name]-1`, `[ws_name]-2` etc. The names seem to be unique. In the workbench the figures are currently `Figure 1`, `Figure 2` etc. Perhaps the automatic naming should also include the plot type, or spectrum number?

ianbush:
I'll merge this as I think there are no major objections to the design outlined.

PR #61, opened 2018-07-27
Design document for InstrumentRayTracer 2.0

BhuvanBezawada:
**Details**
This PR is for a design document regarding `InstrumentRayTracer 2.0`.
The design document outlines possible implementations and how to go about using `InstrumentRayTracer 2.0`.

**Issue** 
mantidproject/mantid#23005

### DIFF
--- a/Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
+++ b/Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
@@ -2,24 +2,9 @@
 ## Design Document
 
 ## Introduction
-The current implementation of [`InstrumentRayTracer`](https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/inc/MantidGeometry/Objects/InstrumentRayTracer.h#L56) works off the orginal `Instrument` API and not the `Instrument 2.0` layers, such as `ComponentInfo`, `DetectorInfo` and `SpectrumInfo`. Due to this, there is a dependency on `Instrument` in `Peak` and as such it is currently not possible to move away from using this legacy API. The intention is to stop developers needing access to the `Instrument` API and instead steer them towards using the new `Geometry` and `Beamline` layers as mentioned above.
+The current implementation of [`InstrumentRayTracer`](https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/inc/MantidGeometry/Objects/InstrumentRayTracer.h#L56) works off the orginal `Instrument` API and not the `Instrument 2.0` layers, such as `ComponentInfo`, `DetectorInfo` and `SpectrumInfo`. Due to this, there is a dependency on `Instrument` in [`Peak`](https://github.com/mantidproject/mantid/blob/master/Framework/DataObjects/inc/MantidDataObjects/Peak.h#L182) and as such it is currently not possible to move away from using this legacy API. The intention is to stop developers needing access to the `Instrument` API and instead steer them towards using the new `Geometry` and `Beamline` layers as mentioned above.
 
-## Overview
-The goal of this document is to outline a plan for the new `InstrumentRayTracer` and to figure out how to move away from making any calls to the legacy API. The areas of discussion in this document include:
-
- * Implementation details of `InstrumentRayTracer 2.0`
-   * Reuse of original code
-   * Possible new methods
-   * Usage of new implementation
-     
- * Using the functionality of `ComponentInfo`
-   * Determining what code would actually be useful
-   * Incorporating it into the new implementation
-   
- * Rolling out the changes
-   * Figuring out what a possible API would look like 
-   * Outlining what changes would have to be made to the code base
-   * How to ensure that performance is good enough
+The goal of this document is to outline a plan for `InstrumentRayTracer 2.0` and to figure out how to move away from making any calls to the legacy API.
  
 ## Implementation Details of `InstrumentRayTracer 2.0`
 
@@ -33,32 +18,23 @@ One idea for the implementation of `InstrumentRayTracer 2.0` is to have many of 
  * `void fireRay(Track &testRay) const`
 
 #### Possible New Methods
-Although the headers do not explicity have anything to do with `Instrument`, many of the method bodies make use of the member variable `m_instrument` to carry out various tasks. This variable is created and initialised in the constructor. From looking at the implementation in [`InstrumentRayTracer.cpp`](), it seems that the constructor is the one place where there is a dependence on `Instrument` and this is probably the one area that needs to be fully rewritten.
+Although the headers do not explicity have anything to do with `Instrument`, many of the method bodies make use of the member variable `m_instrument` to carry out various tasks. This variable is created and initialised in the constructor. From looking at the implementation in [`InstrumentRayTracer.cpp`](https://github.com/BhuvanBezawada/mantid/blob/master/Framework/Geometry/src/Objects/InstrumentRayTracer.cpp), it seems that the constructor is the one place where there is a dependence on `Instrument` and this is probably the one area that needs to be fully rewritten. To rewrite the constructor it may be enough to pass in a `const` reference to a `ComponentInfo` object.
 
-#### Usage of New Implementation
-If all goes well with the plan to use `ComponentInfo` where possible, it is very likely that the code to use `InstrumentRayTracer 2.0` will be very similar to the old way.  
+#### Usage of `ComponentInfo`
+If all goes well with the plan to use `ComponentInfo` where possible, it is very likely that the code to use `InstrumentRayTracer 2.0` will be very similar to the current method of using it. It seems that using the functionality of `ComponentInfo` is very much possible and might well be the best way to proceed.
 
-## Using the Functionality of `ComponentInfo`
-It seems that using the functionality of `ComponentInfo` is very much possible and might well be the best way to proceed.
-The newly exposed `ComponentInfo` layer is accessible from a workspace - examples of usage can be found [here](). 
-
-#### Useful Code
-Possible methods that could be used in `InstrumentRayTracer 2.0` might include:
+Possible methods from `ComponentInfo` that could be used in `InstrumentRayTracer 2.0` might include:
 
  * `componentInfo.getSource()` in `void InstrumentRayTracer::trace(const V3D &dir) const`
  * `componentInfo.getSample()` in `void InstrumentRayTracer::traceFromSample(const V3D &dir) const`
 
-#### Incorporating Code
-It should not be too difficult to reuse the code from `ComponentInfo`. It should just be a case of making sure the `ComponentInfo` class is accessible from `InstrumentRayTracer 2.0`.
+It should not be too difficult to reuse the code from `ComponentInfo` as it should just be a case of making sure the `ComponentInfo` class is accessible from `InstrumentRayTracer 2.0`.
 
 ## Rolling out the Changes
 Before rolling out any changes, it is probably a good idea to develop a new set of tests exclusively for this new class. Some tests (especially for methods that do not make calls to the legacy API) could probably be based on the existing tests with a few alterations. 
-
-#### Figuring out a Possible API
-Once the `InstrumentRayTracer 2.0` class is actually implemented, it might be worth creating a separate design document (as done for `ComponentInfo`, `DetectorInfo` and `SpectrumInfo`) to breifly outline which methods would be best to expose to the Python side.
 
 #### Changes to the Code Base
 The best way to completely move away from using `InstrumentRayTracer` (and not completely break everything) would be to start with a selection of "testing files" that currently use the `InstrumentRayTracer` e.g `Peak` and begin to change them one by one. Once all of those files are in working order it would provide enough confidence in the new implementation and all the other files can then be updated.
 
 #### Performance Issues
-To ensure that the new implementation performs just as well as the existing version, benchmarking tools could be used to record the current performance and this can be used as a target for the new implementation.
+To ensure that the new implementation performs just as well as the existing version, benchmarking tools could be used to record the current performance and this can be used as a target for the new implementation. It does not seem that there would be any major changes to performance since much of the code should remain somewhat similar. If anything, using `ComponentInfo` should reduce the number of function calls being made. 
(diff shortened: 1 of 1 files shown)


Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
```diff
@@ -0,0 +1,68 @@
+# Instrument Ray Tracer 2.0
+## Design Document
+
+## Introduction
+The current implementation of [`InstrumentRayTracer`](https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/inc/MantidGeometry/Objects/InstrumentRayTracer.h#L56) works off the orginal `Instrument` API and not the `Instrument 2.0` layers, such as `ComponentInfo`, `DetectorInfo` and `SpectrumInfo`. Due to this, there is a dependency on `Instrument` in [`Peak`](https://github.com/mantidproject/mantid/blob/master/Framework/DataObjects/inc/MantidDataObjects/Peak.h#L182) and as such it is currently not possible to move away from using this legacy API. The intention is to stop developers needing access to the `Instrument` API and instead steer them towards using the new `Geometry` and `Beamline` layers as mentioned above.
```
OwenArnold: A bit more on the motivation. Read performance via Instrument 2.0 is much better.

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
```diff
@@ -0,0 +1,68 @@
...
+
+#### Performance Issues
+To ensure that the new implementation performs just as well as the existing version, benchmarking tools could be used to record the current performance and this can be used as a target for the new implementation. It does not seem that there would be any major changes to performance since much of the code should remain somewhat similar. If anything, using `ComponentInfo` should reduce the number of function calls being made. 
+
+## Conclusion
+Based on the options outlined above, it seems that the best approach may be to make `InstrumentRayTracer 2.0` a class based implementation. The constructor should take in a copy of a `ComponentInfo` object. The reasoning behind this is that the current implementation is class based and keeping the new version familiar in design might be good for end users. Also, there should not be any extra overhead in terms of performance since the implementation strategy is similar to the previous version.
```
OwenArnold: I disagree with this. Say you wanted to trace a ray. With the class approach I need to write 
```cpp
InstrumentRayTracer instrumentRayTracer(ws.componentInfo());
instrumentRayTracer.trace(V3D{0,0,0});
auto links = instrumentRayTracer.getResults();
```
End users (read Mantid Developers) would likely find the following one-liner more convenient,
```cpp
auto links = InstrumentRayTracer::trace(ws.componentInfo(), V3D{0,0,0});
```
Note that these are slightly different considerations from our Python API, which we would put more effort into preserving backwards compatibility. That restriction does not exist for our C++ API especially for internals like the `InstrumentRayTracer`

For performance reasons, I also think the latter would be better.
1. No copy of `ComponentInfo` required, even though the COW mechanism makes copies of `ComponentInfo` relatively cheap, this is unnecessary.
1. `trace` internally calls `reset` on on `Track` (`m_resultsTrack`) every call, it would likely be just as efficient to construct a new `Track` every time.
1. `getResults` returns by value and copies the links into a new temporary before resetting/clearing it's internal state.

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Rolling out the Changes
+Before rolling out any changes, it is probably a good idea to develop a set of tests exclusively for this new class. Some tests (especially for methods that do not make calls to the legacy API) could probably be based on the existing tests with a few alterations. 
+
+#### Changes to the Code Base
+The best way to completely move away from using `InstrumentRayTracer` (and not completely break everything) would be to start with a selection of "testing files" that currently use the `InstrumentRayTracer` e.g `Peak` and begin to change them one by one. Once all of those files are in working order it would provide enough confidence in the new implementation and all the other files can then be updated.
+
+#### Performance Issues
+To ensure that the new implementation performs just as well as the existing version, benchmarking tools could be used to record the current performance and this can be used as a target for the new implementation. It does not seem that there would be any major changes to performance since much of the code should remain somewhat similar. If anything, using `ComponentInfo` should reduce the number of function calls being made.
```
OwenArnold: Agreed. I think performance tests already exist see `InstrumentRayTracerTestPerformance` in `InstrumentRayTracerTest.h` you should be able to run this now.

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
```diff
@@ -0,0 +1,68 @@
...
+
+#### Classless Implementation
+Another idea could be to make the methods of `InstrumentRayTracer 2.0` a set of free functions. Each of the functions would need to take in a `ComponentInfo` and any other variables required to carry out the correct procedure. This approach would definitely eliminate the passing around of the `InstrumentRayTracer` object that currently happens. Also, there are not many variables that would need to be passed to each of the methods meaning calls to the functions would remain largely the same.
+
+## Usage of `ComponentInfo`
+If all goes well with the plan to use `ComponentInfo` where possible, it is very likely that the code to use `InstrumentRayTracer 2.0` will be very similar to the current method of using it. Possible methods from `ComponentInfo` that could be used in `InstrumentRayTracer 2.0` might include:
+
```
OwenArnold: Is this an exhaustive list? This is one thing we need to be definite about. i.e. is it possible to use `ComponentInfo` for the job of `Instrument` without changes to current `ComponentInfo`?

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
```diff
@@ -0,0 +1,68 @@
...
+ * `componentInfo.getSource()` in `void InstrumentRayTracer::trace(const V3D &dir) const`
+ * `componentInfo.getSample()` in `void InstrumentRayTracer::traceFromSample(const V3D &dir) const`
+
+It should not be too difficult to reuse the code from `ComponentInfo` as it should just be a case of making sure the `ComponentInfo` class is accessible from `InstrumentRayTracer 2.0`.
+
+## Rolling out the Changes
+Before rolling out any changes, it is probably a good idea to develop a set of tests exclusively for this new class. Some tests (especially for methods that do not make calls to the legacy API) could probably be based on the existing tests with a few alterations. 
+
+#### Changes to the Code Base
+The best way to completely move away from using `InstrumentRayTracer` (and not completely break everything) would be to start with a selection of "testing files" that currently use the `InstrumentRayTracer` e.g `Peak` and begin to change them one by one. Once all of those files are in working order it would provide enough confidence in the new implementation and all the other files can then be updated.
```
OwenArnold: As explained below. Breaking things is OK as long as we also retrofit those areas. We need to have a list of ALL areas where `InstumentRayTracer` is used though. Can you gather that and add links to the github files somewhere in here.

BhuvanBezawada: Yes, that sounds like a good idea! I'll get the files and link to them in a **Relevant Files** section.

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
BhuvanBezawada: I'll make the list more comprehensive so we have a clearer idea of what we can use.

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
BhuvanBezawada: I see, I think you are right. I think my reasoning was more based on what we would've done for the Python side. It is probably best to use the classless implementation if it's easier for developers and there are performance benefits.

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
BhuvanBezawada: Ah okay I see. So my understanding is that the `InstrumentRayTracer` would be working in real space whereas the `PeakRayTracer` would be working in reciprocal space? If that's the case, would it make sense for the `PeakRayTracer` to be a wrapper around `InstrumentRayTracer`?

OwenArnold:
@BhuvanBezawada can you include plans for how `Peak::m_instrument` replacement would happen?

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
```diff
@@ -0,0 +1,126 @@
...
+ptr.trace(V3D(0,0,0));
+
+// With the following scoping, there would be undefined behaviour
+InstrumentRayTracer2* ptr;
+{
+  auto ws = CreateWorkspace();
+  ptr = new InstrumentRayTracer2(ws.componentInfo);
+}
+
+ptr.trace(V3D(0,0,0));
```
OwenArnold: This is certain undefined territory. const refs only extend the lifetime of local temporary objects. Does not apply to members.

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
VickieLynch: Unless we are using MDLeanEvents, all the MDEvents have their detectorID.  Instead of ray tracing the box center back to the detector, why don't we use the detectorID for some event near the center or use statistics on all the detectorIDs of events in the box?

OwenArnold: I think that would be an additional feature/change that is outside the scope of what Bhuvan is considering here. Bhuvan is attempting to factor out the `Instrument` object from `Peak` in this first pass. `ComponentInfo` still has the ability to deal with detectorID, so can do exactly what `Instrument` currently does, so the changes would still allow you to develop what you suggest Vickie.

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
```diff
@@ -68,6 +68,9 @@ Before rolling out any changes, it is probably a good idea to develop a set of t
 #### Changes to the Code Base
 The best way to completely move away from using `InstrumentRayTracer` (and not completely break everything) would be to start with a selection of "testing files" that currently use the `InstrumentRayTracer` e.g `Peak` and begin to change them one by one. Once all of those files are in working order it would provide enough confidence in the new implementation and all the other files can then be updated.
 
+#### Changes to Peak
+`Peak.h` has a member variable called `m_inst`. This is the kind of thing we would like to get away from using. By using `ComponentInfo` we can replace almost all the calls made using the `m_inst` variable. However, passing in a `const ComponentInfo` reference may end up being a bit expensive due to the copying of the shared pointers that a `ComponentInfo` holds. One way to avoid this extra expense is to pass in an `ExperimentInfo` object which has access to the `ComponentInfo` object required to make the necessary calls. This method should be cheaper and just as effective. If the classless implementation option is chosen, there would be no need to even store the `m_inst` variable.
```
OwenArnold: You mean an `ExperimentInfo` `const` `shared_ptr`! We don't want to copy an `ExperimentInfo`

BhuvanBezawada: Ah yes of course - I'll fix that!

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
AndreiSavici: Note for MDEvents you also have an experiment info per event. You might have events corresponding to different experiment infos. It is possible that the instrument geometry changes as a function of some logs, so that needs to be taken into account

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
samueljackson92: @BhuvanBezawada Yes that is how I'd imagine it would work.

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
samueljackson92: I agree. I prefer the latter two implementation suggestions. 

I think both suggestions have their strengths. The `ComponentInfo` implementation is nice in the case that you wish to trace multiple times for the same instrument. That implementation removes the need for the `ComponentInfo` object to be in scope for every trace call. 

However the classless implementation makes it very explicit what set of components you're tracing on. This would also make the usage simpler: the most common usage is to trace to get a detector. Currently this is two function calls. With a classless implementation I guess this would drop to one. Something like this?

```cpp
auto detector = InstrumentRayTracer2::traceDetector(compInfo, direction);
```

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
```diff
@@ -0,0 +1,126 @@
...
+
+#### Changes to the Code Base
+The best way to completely move away from using `InstrumentRayTracer` (and not completely break everything) would be to start with a selection of "testing files" that currently use the `InstrumentRayTracer` e.g `Peak` and begin to change them one by one. Once all of those files are in working order it would provide enough confidence in the new implementation and all the other files can then be updated.
+
+#### Changes to Peak
+`Peak.h` has a member variable called `m_inst`. This is the kind of thing we would like to get away from using. By using `ComponentInfo` we can replace almost all the calls made using the `m_inst` variable. However, passing in a `const ComponentInfo` reference may end up being a bit expensive due to the copying of the shared pointers that a `ComponentInfo` holds. One way to avoid this extra expense is to pass in a `const ExperimentInfo` shared pointer object which has access to the `ComponentInfo` object required to make the necessary calls. This method should be cheaper and just as effective. If the classless implementation option is chosen, there would be no need to even store the `m_inst` variable.
```
samueljackson92: I agree with this. I don't think peak should necessarily hold a reference to an instrument. Refactoring to use `ExperimentInfo` seems like a logical step.

Design/Instrument-2.0/Instrument_Ray_Tracer_2.md
```diff
@@ -69,7 +69,7 @@ Before rolling out any changes, it is probably a good idea to develop a set of t
...
```
OwenArnold: So we need to survey to check that it is feasible to get hold of the `ExperimentInfo` at all points where we create a `Peak`, and establish what changes would be required (for example a factory creating Peaks for a given `Instrument` would need to be changed to hold the `ExperimentInfo` instead). Maybe @samueljackson92 could help with this once you have your list of places where `Peak` is made.

OwenArnold:
@samueljackson92 Are you happy with this - shall we get this merged?

samueljackson92:
Yes let's.

PR #65, opened 2018-11-13
Project Save on Mantid Workbench Proposal

Pasarus:
This is a proposal for project save on Mantid Workbench.


Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,115 @@
...
+
+**Both C++ and Python**
+- Our own method of serialization utilizing a JSON back-end
+	- Pros: We can pass C++ values to Python based serializer and back to C++, we can determine file format, we can write custom conversions for types (Most Qt types have a QVariant type conversion to JSON anyway), highly flexible
+	- Cons: Requires extra work to implement that a already available library 
+
+
+Implementation of save window, the recommended Method
+-------------------------------------------------------------------------
+Utilizing the serializing method with a JSON back-end, will likely be best as not only will it use a standard file type, it should be easy to change the file type by changing a few functions inside the serializer. Each UI that is serializable should have an X and Y function where X would create a Dictionary with a string as the key and the value set to whatever object you would need to save. Y would use the string key to get back X's value and emplace the original value again, and update the GUI to represent these values.
```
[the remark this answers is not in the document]
Pasarus: The reason I used X and Y is because I had yet to pick a name and wanted some feedback, the issue being that save and load make it sound like it does the actual work of saving and loading and isn't just being called by the save and loading algorithms etc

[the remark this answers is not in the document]
samueljackson92: > save and load make it sound like it does the actual work of saving and loading and isn't just being called by the save and loading algorithms

Sounds reasonable, but this would be a good document to suggest a name in. What would you call it instead?

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,115 @@
...
+The reason for the split is so that it can be handled in a more abstract manor. Instead of thinking of the task as one great big issue it can be split into the two smaller problems and then split further from there, later on in design.
+
+This document exists because Project Saving is a non-trivial challenge to tackle when it comes to ideas with saving and loading interfaces created in both Python and C++ as well as saving and loading of workspaces. This issue should be fairly simple to understand, simply users should be able to save their projects and load them back in, to at least the standard that MantidPlot was able to achieve.
+
+Changes to MantidWorkbench needed
+-----------------------------------------------
+- MainWindow should track all of the open windows (as well as if they are serializable or not) similar to MantidPlot
+- File->Save Project needs to be added
+- File->Load Project needs to be added
+- A GUI for picking where to Save
```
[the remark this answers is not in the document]
Pasarus: Yes and no, Currently I have been focusing on the back-end of the feature and less on GUI design and implementation the current idea is to save the project as a whole and certainly give support for a single click button before popping up a more advanced GUI for individual saving support, this is alongside support that is mentioned further down for not needing to overwrite already saved workspaces. I shall add a note somewhere about the advanced and not so advanced sections.

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,115 @@
...
+}
+```
+The way this will work alongside the Serialization is that it will all be saved in the same folder, this is the same as the previous implementation (in MantidPlot), for saving and loading workspaces. However a key differences are that this will be produced in Python as that is what Workbench has been produced in, and the output folder will consist of .nxs workspaces and a .json file with the details of the windows.
+
+A workspace should only save over an already saved version of that workspace if it has changed since last save. This can be done by checking the last modified of any files with the same name as a workspace's name and then comparing it to that workspace's history for it's most recent entry.
+
+Diagram to display functionality
+----------------------------------------
+Saving starts from the GUI, and from there it passes on the selected options to the Project Saver which will in turn delegate tasks to saving workspaces and windows. The output of workspace saver and window saver are both put into the output folder.
+![Project Flow](file:project-saveflow.png)
```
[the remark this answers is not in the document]
Pasarus: I shall fix this

Design/Workbench/design-projectsaving.md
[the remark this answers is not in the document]
samueljackson92: Yea, its probably a good idea to keep it in mind. So long as the end implementation has a concept of what "save this list of stuff" then it's fairly simple to filter what is not needed before kicking off the serialisation.

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,115 @@
...
+```python
+def json_line_from_dict(dict):  
+    import json  
+    return json.dumps(dict, indent=4) 
+ ```
+
+One potential implementation is to force usage of serialization for current and new GUIs, create a class which inherits from QWidget/QMainWindow (depending on how PyQt/Mantid Workbench implements a window), which includes a basic copy of X and Y from which windows must inherit/must implement the methods of X and Y. That was a simple list of windows can be held by the main application and a call to X and Y made on them, for saving and loading respectively. This implementation will allow for not only other methods to be forced upon GUI designers that may be required but will also guarantee there is a method for project saving.
+
+For the implementation of X and Y, we need to have a copy of the variables that needs to be serialized on the GUI, so for example if a value needs to be serialized then save it just grab the value and load it back in, in the methods, X and Y.
```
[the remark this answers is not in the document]
Pasarus: Yes it does so I have added a bit on it at the end.

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,115 @@
...
+
+Options for Implementation of Workspace Saving
+--------------------------------------------------------
+**Python:**
+- Pickling
+	- Pros: Built in library of Python
+	- Cons: Need to add _ _ getstate_ _ and _ _ setstate_ _ functions to every UI class which must return a pickle-able type (dict, list, int, string etc). Then setstate function. Unconfirmed if possible or not.
+- JSON Encoding
+	- Pros: Built in library of Python, "faster" than Pickling: https://www.benfrederickson.com/dont-pickle-your-data/, Potential to be integrated with C++ as it's universal
+	- Cons: Needs methods in every class
```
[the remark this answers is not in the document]
Pasarus: Yes but, this isn't possible with out an extra library implementation. I should have added that, this is a purely python method as far as I am aware and would leave the C++ interfaces out of it.

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,115 @@
...
+- Serialization with boost 
+	- Pros: We already have it, simple, uses an archive type for template serialization, has forward compatibility when classes change.
+	- Cons: We need to give Friend access to each UI class for the boost::serialize::access, not python.
+- Serialization with cereal
+	- Pros: Very easy to implement using a template of Archive type.
+	- Cons: We don't already have the cereal library, plus we'd need to cite it, as per licensing, not python.
+
+**Both C++ and Python**
+- Our own method of serialization utilizing a JSON back-end
+	- Pros: We can pass C++ values to Python based serializer and back to C++, we can determine file format, we can write custom conversions for types (Most Qt types have a QVariant type conversion to JSON anyway), highly flexible
```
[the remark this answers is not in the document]
Pasarus: Currently all basic python types and all Qt types than can be a QVariant can be encoded easily by the JSON encoder, I have however added a restriction that it should only be serializing dictionaries and thus anything inside it needs a serializable method in the encoder such as stated here: https://docs.python.org/2/library/json.html
It is my belief that you can redefine default(o) and it should allow you to encode any type as long as the person who wishes to encode this new type uses it.
An example is here:
https://www.programcreek.com/python/example/576/json.JSONEncoder

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,115 @@
+﻿
+Project Saving for Mantid Workbench
+==============================
+Motivation
+--------------
+In MantidPlot it was possible to save all currently open windows and all workspaces, and then be able to load then back in the exact(-ish) state that they were in previously. To achieve this goal again in Mantid Workbench, the problem will be split into two where window serialization and saving and therefore reopening will be separate in implementation, from saving workspaces.
+
+The reason for the split is so that it can be handled in a more abstract manor. Instead of thinking of the task as one great big issue it can be split into the two smaller problems and then split further from there, later on in design.
+
+This document exists because Project Saving is a non-trivial challenge to tackle when it comes to ideas with saving and loading interfaces created in both Python and C++ as well as saving and loading of workspaces. This issue should be fairly simple to understand, simply users should be able to save their projects and load them back in, to at least the standard that MantidPlot was able to achieve.
```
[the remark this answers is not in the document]
Pasarus: Yes that makes perfect sense to be present here

Design/Workbench/design-projectsaving.md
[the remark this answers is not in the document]
Pasarus: I've documented this now

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,115 @@
...
+```
+An Example Y:
+```python
+def Y(self, dict):  
+    self.ui.spinBox.setValue(dict["spinBox"]) 
+    self.ui.lineEdit.setValue(dict["lineEdit"]) 
+    self.ui.dateEdit.setValue(dict["dateEdit"]) 
+    self.ui.radioButton.setValue(dict["radioButton"]) 
+    self.ui.checkBox.setValue(dict["checkBox"])
+    self.ui.label_7.setVisible(dict["label_7Visible"])
```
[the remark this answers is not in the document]
Pasarus: We should probably talk about this tommorrow

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,115 @@
...
+
+The reason for the split is so that it can be handled in a more abstract manor. Instead of thinking of the task as one great big issue it can be split into the two smaller problems and then split further from there, later on in design.
+
+This document exists because Project Saving is a non-trivial challenge to tackle when it comes to ideas with saving and loading interfaces created in both Python and C++ as well as saving and loading of workspaces. This issue should be fairly simple to understand, simply users should be able to save their projects and load them back in, to at least the standard that MantidPlot was able to achieve.
+
+Changes to MantidWorkbench needed
+-----------------------------------------------
+- MainWindow should track all of the open windows (as well as if they are serializable or not) similar to MantidPlot
```
martyngigg: I feel a bit like this is making the workbench head back in the direction of `MantidPlot` where a lot of information from application window was required to know what could be serialized. I agree we need access to the various widgets/windows from `mainwindow` but can we leave it up to them to decide what to do when asked to save/load.
The workbench currently uses widgets from the `mantidqt` package through a [plugin mechanism](https://github.com/mantidproject/mantid/blob/master/qt/applications/workbench/workbench/plugins/base.py). This allows the standalone widgets to exist and be resuable by other applications while keeping workbench-specific behaviour inside the workbench.

Design/Workbench/design-projectsaving.md
martyngigg: I'm not sure how we get away from implementing things for each type we want to save?

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,115 @@
...
+
+A basic example of how a line of JSON that may be generated from an object that uses the X function's output, the output is passed as the variable dict. Using indent=4 makes it a more human readable form when actually written out than with no indent as it would be on a single line:
+```python
+def json_line_from_dict(dict):  
+    import json  
+    return json.dumps(dict, indent=4) 
+ ```
+
+One potential implementation is to force usage of serialization for current and new GUIs, create a class which inherits from QWidget/QMainWindow (depending on how PyQt/Mantid Workbench implements a window), which includes a basic copy of X and Y from which windows must inherit/must implement the methods of X and Y. That was a simple list of windows can be held by the main application and a call to X and Y made on them, for saving and loading respectively. This implementation will allow for not only other methods to be forced upon GUI designers that may be required but will also guarantee there is a method for project saving.
```
martyngigg: I think this is basically what we want. Every window that wants to save/load will need to write code to support itself. I would make the class standalone, basically a class to define an interface for others to implement.

There will be a complication for reading as the window/widget won't yet exist to call this method and it would be nice not to have to mandate default constructors.

Design/Workbench/design-projectsaving.md
martyngigg: I think you're saying that each component should only receive the state it requires to recreate itself? I agree with this.
I don't quite understand the part about saving and reloading though...

There is also a question of how we map from sections of the file, presumable string names of widgets, to actual classes and making sure this is robust over time. We might need to define some type of version...

Design/Workbench/design-projectsaving.md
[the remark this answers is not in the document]
martyngigg: Agreed, this is the time to hammer out names etc

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,173 @@
...
+
+The reason for the split is so that it can be handled in a more abstract manor. Instead of thinking of the task as one great big issue it can be split into the two smaller problems and then split further from there, later on in design.
+
+This document exists because Project Saving is a non-trivial challenge to tackle when it comes to ideas with saving and loading interfaces created in both Python and C++ as well as saving and loading of workspaces. This issue should be fairly simple to understand, simply users should be able to save their projects and load them back in, to at least the standard that MantidPlot was able to achieve.
+
+It should as a requirement allow people to share projects across platforms and seamlessly by zipping and emailing or transferring by different methods. This however should not come in the way of pure functionality.
```
OwenArnold: I'm not clear by what you mean by this paragraph - Are you saying this is a functional requirement, but you are setting a lower priority on it?

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,173 @@
...
+---------------------------------
+- MainWindow should track all of the open windows (as well as if they are serializable or not) similar to MantidPlot
+- File->Save Project needs to be added
+- File->Load Project needs to be added
+- A GUI for picking where to Save
+
+GUI and integration initial ideas
+---------------------------------
+It is true that this document is mostly focused on the actual implementation of saving the project and they intricacies of how that would work. However, it is important that this is discussed to some extent before hand. The intial thought process is that you would be able to do both a basic SaveAll and Save Advanced. SaveAll will run through a pretty simple Save everything as long as it's not overwriting workspaces. Save Advanced would allow the users to Select workspaces and windows individually, also adding an option as to whether or not they want to overwrite the workspaces that may or may be saved which is so at worst case users can save everything from scratch.
+
```
OwenArnold: Fo you see the `Save Advance` as a per-save setting. As in the user would be prompted every time to state exactly what items they wanted to save? Is there any thoughts towards a general persisted set of advanced settings?

Design/Workbench/design-projectsaving.md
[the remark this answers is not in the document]
OwenArnold: I agree with @samueljackson92 about keeping serialisation/de-serialisation away from the main class.

Design/Workbench/design-projectsaving.md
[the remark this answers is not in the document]
Pasarus: Perhaps serialize and de-serialize? Although I don't know whether that would be particularly useful as a name because it isn't particularly clear what it will be used for by just glancing at it, but it seems like a more pertinent name considering it must return a dictionary for the documentation to work and isn't really handling the save and/or loading of anything.

Design/Workbench/design-projectsaving.md
Pasarus: It should be aimed for, but if there is no way around it that would sacrifice ability to actually save then ignore it.

Design/Workbench/design-projectsaving.md
Pasarus: I hadn't given that much thought but it should be possible with QSettings/Serialization I presume.

Design/Workbench/design-projectsaving.md
martyngigg: I disagree. I think it should be a hard requirement that they are portable.

martyngigg: I think we maybe need a section here actually of bullet-pointed requirements at the top.

OwenArnold: @martyngigg yes, it would very good to have that.

Design/Workbench/design-projectsaving.md
Pasarus: Instead we should probably track this in the plugin mechanism you linked. Thus still allowing us to know what is open and what isn't etc. It doesn't have to be in that central place.

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,229 @@
...
+Options for Implementation of Workspace Saving
```
martyngigg: Could we remove this entirely. I think offering the possibility of changing the workspace save mechanism is well outside of the scope here and this slightly suggests that it's worth thinking about. Maybe just have this section say we will use the existing workspace saving mechansims?

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,229 @@
...
+  if "__object_type__" in dict:
+    # Return an object of the original object type
+    return object(dict["object_part1"], dict["object_part2"], dict["object_part3"])
+  return dict
+```
+
+**C++**
+Similarly with the C++ area of serialisation exposing most of the required details to python will be the hardest part, however if most of the interface's details are exposed properly then the decoder and encoders should be able to work exactly the same on these interfaces if only with a little extra implementation on the UI porter/maintainer's part.
+
+Options for Implementation of Workspace Saving
```
martyngigg: I think this is duplicated from the section above?

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,229 @@
...
+	- Cons: We don't already have the cereal library, plus we'd need to cite it, as per licensing, not python.
+
+**Both C++ and Python**
+- Our own method of serialization utilizing a JSON back-end
+	- Pros: We can pass C++ values to Python based serializer and back to C++, we can determine file format, we can write custom conversions for types (Most Qt types have a QVariant type conversion to JSON anyway), highly flexible
+	- Cons: Requires extra work to implement that a already available library 
+
+
+Implementation of save window, the recommended Method
+-----------------------------------------------------
```
martyngigg: I'm not so sure about this section - it's not entirely clear how all of the bits fit together. I think a few key points as bullets would help first like:
* an intermediary format will be used for windows to save/load project settings -> each window doesn't have to know about JSON for example. A `dict` is the most convenient type for this.
* for each type that will be saved to a project a pair of classes will be created, an Encoder and a Decoder
  * the first is responsible for creating a `dict` of the stored settings
  * the second is responsible for create the type from a `dict` of settings
* separate types for encoding/decoding rather than methods on each saveable type have advantages:
 * we can't have additional methods on `matplotlib` plots so separate classes is the only way there
 * we can write all project saving in python, even for C++ widgets
 * unit testing of the separate pieces will be more effective and less coupled together

I like the sequence diagram from below and maybe that should then follow this at the top of this section with then a paragraphs explaining each class in the sequence diagram?

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,229 @@
...
+Utilizing the serializing method with a JSON back-end, will likely be best as not only will it use a standard file type, it should be easy to change the file type by changing a few functions inside the serializer. Each UI that is serializable should have a encode and decode ability where encode would create a dictionary with a string as the key and the value set to whatever object you would need to save. Decode would use the string key to get back encoder's value and emplace the original value again, and update the GUI to represent these values.
+
+**Saving:**
+Utlizing an abstraction and encapsulation method we have a few major sections to tackle with a middle man "Saver Class". To save we would have the section which encompasses encoding. These sections are encoding, clumping encoded pieces together ("Saver") and writing, similar to how MVP works with GUIs. This way the encoding and writing parts will not need to touch eachother and can be swapped out independently as well as tested seperately with mocking and unit tests.
```
martyngigg: The encoder/decoder classes seem to come a bit out of nowhere here. I think moving the sequence diagram above will help.

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,229 @@
...
+
+For encoding, largely this will be left up to interface implementers and will be fairly simple to do, either by hand or a more automatic approach (a more automatic approach has the unintended consequence of not allowing backwards compaitibility easily when changing variable or widget names). The proposed idea is for developers to create a InterfaceXEncoder Class where InterfaceX is the name of the Interface. This InterfaceXEncoder will be availible to the EncoderFactory which will allow developers to refactor code and be able to still load older data.
```
martyngigg: I'm not quite sure I agree that the "encoding will be left up to interface implementers". We want them to create the encoder/decoder types, right? Do you mean the method inside the encoder/decoder are up to interface implementers?

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,229 @@
...
+class exampleDecoder():
+  def decode(self, dict):
+    TestUI(dict["arg1"], dict["args2"])
+    TestUI.ui.spinBox.setValue(dict["spinBox"]) 
+    TestUI.ui.lineEdit.setValue(dict["lineEdit"]) 
+    TestUI.ui.dateEdit.setValue(dict["dateEdit"]) 
+    TestUI.ui.radioButton.setValue(dict["radioButton"]) 
+    TestUI.ui.checkBox.setValue(dict["checkBox"])
+    TestUI.ui.label_7.setVisible(dict["label_7Visible"])
+    TestUI.show()
```
martyngigg: I would leave out `.show` and return the TestUI object to the caller to decide when best to do that.

Pasarus: Agreed :+1:

Design/Workbench/design-projectsaving.md
Pasarus: Yes that is what was intended, I have added a clearer implementation.

Design/Workbench/design-projectsaving.md
Pasarus: Done :+1:

Design/Workbench/design-projectsaving.md
Pasarus: Agreed and Done :+1:

Design/Workbench/design-projectsaving.md
Pasarus: Not quite, but I will expand the title of this subsection to explain a little better what it is

Design/Workbench/design-projectsaving.md
Pasarus: Removed the vast majority of it but I will leave a bit to explain what to do with workspaces.

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,203 @@
...
+    return object(dict["object_part1"], dict["object_part2"], dict["object_part3"])
+  return dict
+```
+
+**C++ Saving and Loading Integration:**
+Similarly with the C++ area of serialisation exposing most of the required details to python will be the hardest part, however if most of the interface's details are exposed properly then the decoder and encoders should be able to work exactly the same on these interfaces if only with a little extra implementation on the UI porter/maintainer's part.
+
+Implementation of save workspaces and linking to saved scripts
+--------------------------------------------------------------
+Utilizing the already implemented SaveNexusProcessed and SaveMD algorithms, the suggestion is to implement a python version of how MantidPlot currently implements saving workspaces. Call SaveNexusProcessed for most workspaces and SaveMD for EventWorkspaces. Saving a copy of the names of the saved workspace .nxs files in a JSON format in the same file as the output from the serialization of the windows, it may look like this:
```
soininen: Please note that `SaveNexusProcessed` does not work on `MaskWorkspace`s and thus our current project saving does not save/restore such workspaces. One needs to use `SaveMask` instead.

Design/Workbench/design-projectsaving.md
```diff
@@ -0,0 +1,203 @@
...
+In MantidPlot it was possible to save all currently open windows and all workspaces, and then be able to load then back in the exact(-ish) state that they were in previously. To achieve this goal again in Mantid Workbench, the problem will be split into two where window serialization and saving and therefore reopening will be separate in implementation, from saving workspaces.
+
+The reason for the split is so that it can be handled in a more abstract manor. Instead of thinking of the task as one great big issue it can be split into the two smaller problems and then split further from there, later on in design.
+
+This document exists because Project Saving is a non-trivial challenge to tackle when it comes to ideas with saving and loading interfaces created in both Python and C++ as well as saving and loading of workspaces. This issue should be fairly simple to understand, simply users should be able to save their projects and load them back in, to at least the standard that MantidPlot was able to achieve.
+
+
+Functional Requirements
+-----------------------
+- Graceful handling of other versions that have been saved previously
```
soininen: I hope I read this right: the saved projects are properly versioned and can be restored by any future version of Mantid.

Pasarus: Not necessarily versioned per say, because anytime a new encoder/decoder would be added or changed it would technically change the version, but the idea is that it should gracefully handle loading an older file.

Design/Workbench/design-projectsaving.md
Pasarus: Noted and will update this document accordingly. The caveat for changing workspace type is already in place of EventWorkspaces so it would just be adding another one to list :+1:

Design/Workbench/design-projectsaving.md
```diff
@@ -175,7 +175,7 @@ Similarly with the C++ area of serialisation exposing most of the required detai
 
 Implementation of save workspaces and linking to saved scripts
 --------------------------------------------------------------
-Utilizing the already implemented SaveNexusProcessed and SaveMD algorithms, the suggestion is to implement a python version of how MantidPlot currently implements saving workspaces. Call SaveNexusProcessed for most workspaces and SaveMD for EventWorkspaces. Saving a copy of the names of the saved workspace .nxs files in a JSON format in the same file as the output from the serialization of the windows, it may look like this:
+Utilizing the already implemented SaveNexusProcessed, SaveMD, and SaveMask algorithms, the suggestion is to implement a python version of how MantidPlot currently implements saving workspaces. Call SaveNexusProcessed for most workspaces, SaveMD for EventWorkspaces, and SaveMask for MaskedWorkspaces. Saving a copy of the names of the saved workspace .nxs files in a JSON format in the same file as the output from the serialization of the windows, it may look like this:
```
martyngigg: I don't think we should make a distinction for `MaskWorkspace`. If it's a workspace then we should just use `SaveNexusProcessed`.

martyngigg: I agree. @SamJenkins1 perhaps we should look at that as a prerequisite?

martyngigg: I agree. @Pasarus perhaps we should look at that as a prerequisite?

martyngigg:
Merging this as it is now implemented in workbench/mantidqt.

PR #66, opened 2018-11-27
Design document for MDWorkspace changes to increase performance.

igudich:
Here is the design document, which describes changes tin the current implementation of MDWorkspace to increase the performance of creating and merging.


SimonHeybrock:
This had been approved and implemented a year ago. Merging.

PR #67, opened 2019-05-22
Trivially Exportable Citations

DanNixon:


Design/trivially_exportable_citations.md
```diff
@@ -0,0 +1,138 @@
...
+  - `description` is always optional (this isn't needed for citation, but gives insight as to why this citation is relevant)
+  - if `bibtex` is provided `endnote` must also be provided, and vice-versa (BibTex and Endnote contain essentially the same information, they can both be created if one can be. BibTex and Endnote do not imply a DOI is minted)
+  - if `doi` is provided, `url`, `bibtex` and `endnote` must all be provided (BibTex and Endnote can be generated from DOIs)
+  - if none of `doi`, `bibtex` or `endnote` are provided, `url` must be provided (there must be something there, even if this isn't citable a URL is better than nothing)
+
+Some helper "toString" methods may be beneficial.
+
+### Algorithm citations
+
+An additional `static` method will be added to `IAlgorithm` that returns a `std::vector<Citation const> const`.
```
martyngigg: Should this say `virtual` similar to things like `version` etc?

Design/trivially_exportable_citations.md
```diff
@@ -0,0 +1,138 @@
...
+### Documentation
+
+There are two main requirements for documentation integration:
+  - Listing the citations in the algorithm documentation page
+  - Allowing the documentation writer (developer) to cite the citations at specific points in the documentation
+
+The first point is trivial as we already do similar things for properties, related algorithms, etc.
+
+The second will involve more work.
+I don't have enough Sphinx know-how to suggest how this may be implemented.
```
martyngigg: Do you want away to arbitrarily reference one of the citations or the whole lot?

Design/trivially_exportable_citations.md
```diff
@@ -0,0 +1,138 @@
...
+## Design considerations
+
+### Property dependant citations
+
+I briefly considered making the citations returned for algorithms dependant on the algorithm being run.
+This allows the algorithm to record what citations it used during execution.
+
+The advantage this gives is greater accuracy of citations (e.g. toggling an option may divert control flow away from a particular citable method).
+
+It was chosen not to go down this route for the following reasons:
```
martyngigg: I agree. I think this is a level of complexity that is not required.

Design/trivially_exportable_citations.md
```diff
@@ -0,0 +1,138 @@
...
+  - It adds complexity
+  - It adds runtime dependency
+    - This prevents tooling for algorithm documentation
+
+### Backwards compatibility
+
+What happens when loading old workspaces with no citations in the NeXus file?
```
martyngigg: I think the only thing we can do is have empty cite fields?

Design/trivially_exportable_citations.md
```diff
@@ -0,0 +1,138 @@
...
+
+### `constexpr`
+
+Is it worth using `char *` as the string format to be able to `constexpr` the `Citation` structure?
```
martyngigg: Maybe better to work out at implementation if we find places that benefit from the information at compile time? I can see though that a lot will just be static strings...

Design/trivially_exportable_citations.md
DanNixon: All of the citations for an algorithm should be listed in the corresponding algorithm documentation.

Ideally there should be a way to reference a citation for an algorithm within the RST documentation for that algorithm (that is the part I can't suggest how to implement here).

I'm not planning to make things citable from anywhere in the documentation, I'm not convinced there is a significant use case for that as you would likely want to cite the algorithm instead.
One exception may be framework specific things (such as MD event implementation).

DanNixon:
I've updated this based on the comments from the meeting, `Citation` can now be created using more typical information (journal, year, authors, title, etc.).

`Citation` itself will still store the same fields as this must mirror `NXcite` to facilitate loading and saving (yes, I also have no idea why `NXcite` doesn't just store the citation information rather than formatted citations).

Design/trivially_exportable_citations.md
```diff
@@ -0,0 +1,183 @@
...
+};
+
+```
+
+In the constructor for the relevant `struct`, `Citation` would then call `to_bibtex` and `to_endmode` functions to obtain citations in the required formats.
+DOI and URL fields can be directly extracted where possible.
+
+As an example, creating a citation could be done like so:
+```cpp
+auto const savu = Citation(PaperCitation{
```
martyngigg: Thanks for the addition of an example. I think removing the requirement for the developer to format the citation properly is a good move.

PR #68, opened 2019-06-20
Discuss how best to introduce the new features

OwenArnold:
Additions to library currently under development. We are a few weeks off looking to expose this at the user level in any form.

Pros and Cons of 3 approaches provided. Would appreciate help in selection from TSC.

### DIFF
--- a/Design/nexusgeometrysaving.md
+++ b/Design/nexusgeometrysaving.md
@@ -1,0 +1,74 @@
+## Overview
+
+Version 4.0 of Mantid saw the introduction of the Nexus Geometry Format [updates](https://docs.mantidproject.org/v4.0.0/release/v4.0.0/framework.html#nexus-geometry-loading) into Mantid. This has gone well with, many of our tests showing speed improvements upon loading for a range of equivalant instruments (Rectangular Detectors excluded) and no loss of functionality or performance for the existing routes.
+
+ESS intend to support the new geometry format for intermediate files because:
+
+1. The format has been demonstrated to suit the representation of neutron instrument geometries well
+1. It is relatively low-cost for the ESS to adopt as part of an intermediate format.
+1. It gives a common format for geometry for Raw and Processed files
+1. It is complete and accurate. The current approach bypasses serialization.
+1. It is not application specific, geometries are useable in downstream applications more easily.
+
+This is a follow up to the Working Group Meeting 20/06/2019. We are aware that ISIS do not intend to make use of the new format. The positions of the SNS and ILL were not represented at that time.
+
+We are currently running a small project involving a student to provide a saving mechanism within the existing the lib [NexusGeometry](https://github.com/mantidproject/mantid/tree/master/Framework/NexusGeometry) to provide a high-level function to support the saving.
+
+**This document is for discussion of the best way to make these features available within Mantid**
+
+## 1. Implicit Replace instrument_xml in `SaveNexusProcessed`
+
+This represents an implicit change in behaviour.
+
+Current mechanism [implemented here](https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Instrument.cpp#L1010), would be replaced (see notes below) to use the new method in `NexusGeometry`. This would mean that ALL new processed nexus files contained the full geometry rather than the xml string. Aligned with that, we would also remove the instrument geometry specific parameters i.e. the things representing overloads to the correct position.
+
+It would be recommended to retain the switch that allowed us to write old-style processed nexus files but the default would be to have the new-style files written out. The switch would be exposed as a parameter on the algorithm.
+
+Reloading the files should work without any modifications, `LoadInstrument` already does the right thing.
+
+**Pros**
+
+1. Users do not need to be aware that this change is happening - they have enough to think about
+1. Provides a useful modernisation update to Mantid across the facilities
+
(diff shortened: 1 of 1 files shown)


Design/nexusgeometrysaving.md
```diff
@@ -0,0 +1,74 @@
...
+
+**Pros**
+
+1. Potential for breaking changes low.
+1. Easy to compare files between versions of Mantid without thinking
+
+**Cons**
+
+1. Users need to be made aware of the switch
+1. Users would not necessarily take advantage of this. More of a pain at facilities like ESS where the default would be to adopt.
```
martyngigg: I agree that if we set the default to not use it feels like we are causing pain for newer facilities and also prolonging the pain for ourselves. Do you think it would be possible to make the default dependent on the instrument's facility? This is a new version of the format and we could have a tag in the `facilities.xml` file to define it?

Design/nexusgeometrysaving.md
```diff
@@ -0,0 +1,74 @@
...
+* No breaking for old scripts
+
+**Cons**
+
+* SaveNexus is a natural name for what we want to do in both cases
+* Takes more time to refactor and create new algorithm
+
+## Notes on the Instrument Service 
+
+The instrument service allows Mantid to cache instruments and prevent reload. Both the `InstrumentDefinitionParser` and the `NexusGeometryParser` have hash mechanism to compare instruments based on content and therefore prevent uncessary reload.
```
martyngigg: This is good and certainly alleviates worries about over using memory for identical copies of things.

Design/nexusgeometrysaving.md
```diff
@@ -0,0 +1,74 @@
...
+
+We are currently running a small project involving a student to provide a saving mechanism within the existing the lib [NexusGeometry](https://github.com/mantidproject/mantid/tree/master/Framework/NexusGeometry) to provide a high-level function to support the saving.
+
+**This document is for discussion of the best way to make these features available within Mantid**
+
+## 1. Implicit Replace instrument_xml in `SaveNexusProcessed`
+
+This represents an implicit change in behaviour.
+
+Current mechanism [implemented here](https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Instrument.cpp#L1010), would be replaced (see notes below) to use the new method in `NexusGeometry`. This would mean that ALL new processed nexus files contained the full geometry rather than the xml string. Aligned with that, we would also remove the instrument geometry specific parameters i.e. the things representing overloads to the correct position.
```
martyngigg: Can I check what happens with non-instrument parameters set in the IDFs? I assume these are still saved/loaded?

Design/nexusgeometrysaving.md
OwenArnold: Yes that could be a solution. Most of the time that would be fine, but I think it could cause annoyance say users trying to understand differences in behaviour (different facilities set but otherwise identical scripts perhaps).

Design/nexusgeometrysaving.md
OwenArnold: Yes, they must be saved still. Though we are not discussing instrument 2.0 here (i don't want to get the Nexus geometry aspects confused with this) it uses a similar process to make parameters that allow current processed nexus files to be writen (Instrument 2.0 also has absolute positions/rotation for all items) see [here](https://github.com/mantidproject/mantid/blob/master/Framework/Geometry/src/Instrument.cpp#L1010)

Design/nexusgeometrysaving.md
NickDraper: Well yes and this will help most of the situations, but not all.

This is because as we seperated the "base" instrument from the alterations (calibration, or movement) then the "base" instrument would be the same and then shared, even if the movements or calibrations were different.

This approach will help for non-moving and non-calibrated instruments, but not for those that move, or have different calibrations.
While most of the time runs in the same Mantid session will be from the same calibration, we will have occaisions when to runs from different cycles are compared, and this could cause significant additional load for the larger instruments.  What will confuse our users more is it will be hard to understand when mantid can and cnanot share the instrument geometry.

OwenArnold: @NickDraper I'm confused by this - I'm not clear on how your comments relate to the file format, you seem to be discussing in-memory "base instruments" here. I'm not clear on how the IDFs do a better/different job than what is being proposed here.

Design/nexusgeometrysaving.md
```diff
@@ -0,0 +1,74 @@
+## Overview
+
+Version 4.0 of Mantid saw the introduction of the Nexus Geometry Format [updates](https://docs.mantidproject.org/v4.0.0/release/v4.0.0/framework.html#nexus-geometry-loading) into Mantid. This has gone well with, many of our tests showing speed improvements upon loading for a range of equivalant instruments (Rectangular Detectors excluded) and no loss of functionality or performance for the existing routes.
+
+ESS intend to support the new geometry format for intermediate files because:
```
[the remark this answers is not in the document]
OwenArnold: 1. Agreed, that is already the case and would require no additional work here.
2. `Any new format must support all of the operations of the old format.`: I'm not clear what you are saying here.
3. `whether a file uses the new or old format should be invisible to the users`: If you mean at the point of serialisation, then we must adopt a route whereby the name remains `SaveNexus`

PR #72, opened 2020-01-23
Add initial reflectometry IDF design doc

giovannidisiena:


Design/reflectometry-idf.md
```diff
@@ -0,0 +1,62 @@
+# Reflectometry IDF
+
+## Motivation
+
+The reflectometry instrument scientists do not currently use an IDF and are heavily reliant on their own scripts.
+If the IDF could properly describe their setup then it would solve existing issues and improve their workflow.
```
OwenArnold: Can you please categorically state all the existing problems that a design needs to address. I don't see how we can consider solutions without a proper definition of the problem. You've listed benefits to solution below, but that's not the same thing.

Design/reflectometry-idf.md
```diff
@@ -0,0 +1,62 @@
...
+
+Potential benefits include, but are not limited to:
+  - Correct geometric description of detector components and calculation of angles to be fed into [algorithms](https://github.com/mantidproject/mantid/issues/26971)
+  - Complex beam paths to allow for accurate description of mirror action, with a view to accurately describe the effect of gravity
```
OwenArnold: Possibly more accurate, though beam paths will still probably just provide an approximation (point to point per component).

Design/reflectometry-idf.md
```diff
@@ -0,0 +1,62 @@
...
+  - Reduce known sources of error due to problems with Q error bars, incorrect angles, gravity corrections, etc.
+  - Facilitate new equipment over multiple instruments/facilities, such as 2D detectors
+
+## Solution
+
+This task should be thought of in terms of four main sub-tasks:
+
+  ### 1. POLREF IDF support for relative pixel map
+  
+  Andrew currently has a scan of the POLREF detector which he uses in a script to move each detector pixel into place every time
```
OwenArnold: Can you elaborate on this scanning technique. What exactly are the steps to this? I think you're saying that they use the specular reflection (along with known incident angle) to identify exactly where the detector is? Do they use [this](https://docs.mantidproject.org/nightly/algorithms/FindReflectometryLines-v2.html) for this or something else?

Design/reflectometry-idf.md
```diff
@@ -0,0 +1,62 @@
...
+  
+  Current workarounds include the following:
+    - Separate 'neutronic' and 'physical instrument' IDFs
+	- Dynamic modification of positions via `MoveInstrumentComponent`
+	- Storage of the L1 & L2 distances in a separate table, which `ConvertUnits` can be made to 'understand'
+	
+  The ideal solution to this task would be to create a new `beam-path` tag for description within the IDF. This can then can be used 
+  with Owen’s [Instrument 2.0 prototype](https://github.com/DMSC-Instrument-Data/instrument-prototype) which specifically deals with 
+  the in-memory side of this task. He has mentioned that ray tracing will not be possible, and even with the prototype discussed above 
+  only one flight path per component has been considered, so additional thought will need to be given to multiple flight paths as this
```
OwenArnold: Yes. That is the limitation. From your description of their procedure above, the complex beam path approach might work for them. Given that they are using L2,L1 tables or neutronic/physical instrument IDFs I think they are already imposing the restriction on themselves of one flight path per component.

Design/reflectometry-idf.md
```diff
@@ -0,0 +1,62 @@
...
+  
+  The instrument scientists would like to describe multiple modes of configuration (NR, PNR liquids, etc.) within a single IDF. 
+  This will require a sensible way to describe multiple degrees of freedom and, if done accurately, will help in solving an issue they 
+  are having with Q error bars and some other possible source of systematic error. It will also allow them to update their workflow in 
+  real time as the report often having to change configurations with some users which ultimately takes a lot of time.
+  
+  A possible solution to this would be to have a generic IDF which can obtain the specified configuration through the log files delivered 
+  by SECI, as is currently done in some IDFs. The IDF would then be able to update accordingly through a switch-statement of sorts, 
+  moving the relevant components into place. No such log currently exists and so if this solution is decided upon it will require
+  collaboration with the IBEX team.
```
OwenArnold: Other solutions based on delivering the geometry with the data:

1. The IDF could potentially be written into the raw nexus file. I believe since LoadInstrument is used, it would automatically process that. This would be a change on the ECP side (SECI) but I prefer it to adding logs as that's just one additional step that could break along the way.
2. Move to use the NexusGeometry Format in the nexus files. This would be a larger change on the SECI /IBEX side but would automatically solve these problems. This should be discussed in conjunction with the NexusConstructor with @matthew-d-jones

Design/reflectometry-idf.md
giovannidisiena: Yes, my current understanding of the issues are the sub-tasks listed below, but I will add a couple of lines to this section as well. I will probably also send this motivation section to the scientists to verify and hopefully fill some of the gaps in the detail.

Design/reflectometry-idf.md
giovannidisiena: Did @gvardany do some work on a similar issue recently? I think it would be best to leave it as a point to point approximation and handle gravity corrections elsewhere but I don't know how that would be implemented. From what I understand, there are sometimes multiple bounces off mirrors before reaching the detector so these will all affect the final angles that are passed into the algorithms.

gvardany: There was an attempt by my colleague to implement a general gravity correction algorithm https://github.com/mantidproject/mantid/pull/26023 
However this is stale and outdated, so trying to recover it would be quite painful.
Besides it has never been properly validated. I'm tempted to presume, that redoing from scratch might actually be faster. Yet it is not at all trivial in case of multiple mirrors...

Design/reflectometry-idf.md
giovannidisiena: I don't have any further information about the scan other than that it is of theta vs channel, and was necessary because they don't know where each pixel is. But yes, from what I can tell they use specular reflection along with a known incident angle to work backwards from the expected reflected angle and move the detector into place (with the relative pixel positions described by the scan/relative pixel map).

Design/reflectometry-idf.md
giovannidisiena: I don't they are using these workarounds. I added them as background after talking to you and Nick about existing solutions that other instruments use, but maybe that is confusing and needs to be made explicit.

Design/reflectometry-idf.md
OwenArnold: It is necessary for you to confirm with the team that that is what they are doing. I.e some simple trig based on incident angle and known sample-detector distances.

OwenArnold:
@giovannidisiena Are you still working on this? Work seems to have stalled over the last 2 months

giovannidisiena:
@OwenArnold somewhat, yes. There was a shift in my priorities but in light of the current situation the scientists are keen to give more focused input, given the extra time on their hands.

Design/reflectometry-idf.md
christykinane: Hi Owen, The scan we are referring to is a detector map, where we work out the position of each pixel in Theta (degrees). We do this by putting a supermirror at the sample point to bounce the beam up. This is just so we can used the bench to map the whole detector as if you use the straight through beam then the detector bench does not drive low enough. We then start with the detector as low as possible with the beam off the top of the detector. The bench is then moved up in very small steps in theta. I calculated a minimum step size via the special Nyquist frequency. We then begin and end a run from some small number of micro amps (we actually used event mode with a high frequency log of the theta angle and sliced it all up later). Then we move by the next small step and do another TOF scan and so on, until we sweep of the bottom of the detector. Once we have this scan, we then go through it for each pixel and obtain as a function of angle the intensity on that pixel. This is effectively a theta scan for that pixel; we fit it and obtain for pixel 1 say its position in theta. We do this for all 640 pixels on the detector. we define 0 theta as pixel 280, but that does not really matter, as its the relative distances between the pixels that’s important, The way the detectors are made means that the pixels are not evenly spaced and this allows use to obtain a pixel position and error on it form the fit. We can use this then to remap the detector map to actual theta positions. This means the data converts to Q correctly for 2D maps. It also allows proper implementation of Q summing, gravity etc. We cannot just use an IDF detector description where we say each pixel is just 0.5mm apart as we get the wrong answer when converting to Q as each pixel is uniquely positioned, because of the manufacturing process. We just load this pixel vs theta data set in (2-column dataset) and use it to calculate relative pixel positions. 

I hope this helps, please ask my or Andrew any questions you have.

PR #78, opened 2020-08-26
Discussion on strategy for managing 3rd-party dependencies after dropping MantidPlot

martyngigg:
As discussed previously we held off doing anything on the dependency management until we dropped MantidPlot. We are now near that point so kicking off this discussion with an overview of the current position and some solutions.

Time to discuss this further and come to an agreement on a strategy.

### DIFF
--- a/Design/ThirdpartyDependencies.md
+++ b/Design/ThirdpartyDependencies.md
@@ -53,7 +53,7 @@ Pros:
 * Smaller user packages. The system libraries are shared for all packages resulting
   in a much smaller package to install.
 * (Mostly) managed for us.
-* Simply commands to install packages (apt/yum)
+* Simple, well-known commands to install packages (apt/yum)
 
 
 Cons:
@@ -91,13 +91,18 @@ Cons:
 * Homegrown set of scripts are very clunky.
 * Maintenance of build scripts adds overhead.
 * Complete bundling of dependencies gives a much larger user installer package.
+* ABI changes require a complete rebuild
+* Initial download (~5GB) takes quite some time
 
 
 ### MacOS
 
-MacOS uses the [Homebrew](https://brew.sh/) package manager along with a custom
-set of [formulae](https://github.com/mantidproject/homebrew-mantid) for obselete
-dependencies.
+MacOS uses the [Homebrew](https://brew.sh/) package manager for C++ dependencies.
+A custom [tap](https://github.com/mantidproject/homebrew-mantid) provides a
+[mantid-developer](https://github.com/mantidproject/homebrew-mantid/blob/master/mantid-developer.rb)
+meta-formula to install all of the dependencies in one shot. 
+Python dependencies are installed with `pip` and a 
+[requirements file](https://github.com/mantidproject/homebrew-mantid/blob/master/requirements.txt).
 Homebrew either downloads a prebuilt binary or builds it locally from the formula.
 A developer formula is provided to easily install the required dependencies.
 
@@ -105,16 +110,20 @@ The package provided for users contains all required dependencies
 much like Windows.
 
 Pros:
-* Build formula (mostly) maintained for us.
-* Simply commands to install packages
+* Formula for dependencies (mostly) maintained for us.
+* Simple commands to install packages
 
 Cons:
 * Dependencies not versioned with the main code. Hard to track versions used.
-* No way to set deployment. We have to build on the earliest OS we intend
+* No way to set deployment target. We have to build on the earliest OS we intend
   to support causing issues with hardware availability.
 * New dependencies have to be installed by hand on developers
   and builders machines.
-* Versions can be pinned but Homebrew really likes to have everything at HEAD
+* Versions can be pinned but Homebrew really likes to have everything at HEAD.
+  This has recently caused issues when Python-based formula, like PyQt, were migrated
+  to Python 3.9 but many other Python dependencies such as matplotlib were
+  not yet available for 3.9. Homebrew has to be pinned to we receive no more updates
+  in this case.
 
 ## General Concerns
 
@@ -126,11 +135,14 @@ there are also general issues with the current approach:
    This can cause confusion for developers
    where their platform might have a newer api
    yet their code does not work on other platforms.
-   This slows down development.
+   This slows down development and ties us to the oldest
+   LTS for years.
 2. Using a new library takes a lot of overhead and
    in some cases is basically impossible
 3. Three different methods requires much additional overhead
    in understanding how things are managed.
+4. Pushing new dependencies to build servers requires
+   co-ordination and potential downtime.
 
 ## Requirements
 
@@ -163,9 +175,15 @@ The solutions considered in detail are:
 
 Other solutions not considered:
 
-* Flatpak/Snap - would only work for Linux and still be a large effort.
-* Singularity/Docker - unsure about desktop applications & would not work for macOS.
-
+* Flatpak/Snap:
+  * Would only work for Linux and still be a large effort.
+* Singularity/Docker:
+  * Unsure about desktop applications & would not work for macOS.
+  * Docker on Windows requires the Hyper-V role, so users would
+    have to get admin access and have a processor capable of 
+    exposing the required instructions - so it's not universally compatible
+* WSL:
+  * Needs to be installed separately. Barrier to high for non-technical users.
 
 ### Conan
 
@@ -218,13 +236,23 @@ that have no other dependencies and are easy to install.
 
 Conan was designed as a framework to package C/C++ dependencies.
 While it is written in Python itself it does not possess the capability
-to manage Python dependencies. Windows/macOS already ship a bundled Python
-distribution so maintaining a separate site-packages is already supported here.
-It is proposed that the same be done for the desktop application bundle
-on Linux so that it is fully self-contained.
+to manage Python dependencies.
+In our current solution on Windows/macOS we already ship a bundled Python
+distribution to users and the proposal is to do the same for linux
+for the workbench application bundle.
 
-The Windows/macOS Python-bundling solution is clunky.
-Another option is creating our own Python Conan package
+Bundling of Python is managed differently for Windows/macOS:
+
+* Windows: The [third party](https://github.com/mantidproject/thirdparty-msvc2015/)
+  repository contains a Python version installed by hand along with 
+  [scripts](https://github.com/mantidproject/thirdparty-msvc2015/blob/master/build-scripts/fixup-python3.bat)
+  to fix up links so the bundle is relocatable.
+  CMake scripts copy the bundle to the Mantid user package.
+* macOS: Homebrew is used to install Python to a dev machine as part
+  of the developer package. A [ruby script](https://github.com/mantidproject/mantid/blob/master/installers/MacInstaller/make_package.rb)
+  copies the bundle to the user package.
+
+Here it is proposed that we create our own Python Conan package
 that can be used like any other dependency for the desktop
 application bundle.
 This has been [prototyped](https://github.com/martyngigg/conan-python3)
@@ -260,12 +288,15 @@ is another cross-platform package manager but with support for a
 variety of languages including Python and C++.
 Alongside `pip` many popular Python packages support distribution through Conda.
 In fact `mantid` already has [Conda packages](https://anaconda.org/mantid/repo)
-for workbench and the framework for Linux and Mac.
+for workbench and the framework for Linux.
+Mac packages are currently effectively abandoned.
 
 [Anaconda](https://repo.continuum.io/pkgs/) hosts many prebuilt packages for
 various combinations of compilers and operating systems.
 [Conda build](https://conda.io/projects/conda-build/en/latest/resources/commands/conda-build.html)
 is provided to enable building custom packages.
+Compilers are somewhat limited., getting latest versions of compilers on all platforms is
+problematic/hacky at best.
 
 The [meta.yaml](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#requirements-section)
 file provides a central place to define dependencies and would satisfy requirements 2-4.
@@ -296,5 +327,9 @@ Questions:
 
   * Where would we store binary artifacts? Or just store recipes?
   * Bintray has been considered but its price could be prohibitive.
+  * What would be the rough time to build locally per machine as an estimate?
+    * If it's an hour per developer per say 6-12 months, configuring remote
+      artefacts / storage / retention vs building locally would become a factor
+
 
 3. Other things not considered here?
--- a/Design/ThirdpartyDependencies.md
+++ b/Design/ThirdpartyDependencies.md
@@ -5,7 +5,7 @@ the current issues surrounding the use of third-party libraries
 within the mantid codebase,
 along with proposals for solutions to the current problems.
 
-# Motivation
+## Motivation
 
 The ESS will be using features from HDF5 such as variable-length strings,
 [single writer multiple reader](https://support.hdfgroup.org/HDF5/docNewFeatures/NewFeaturesSwmrDocs.html)
@@ -17,7 +17,7 @@ is not a good idea as it could lead to a mess of package conflicts.
 Furthermore, while this is an issue with a single package it points to
 a more systemic issue with how we handle our dependencies across the project.
 
-# Current Situation
+## Current Situation
 
 Mantid uses several third-party libraries
 in order to avoid reinventing the wheel,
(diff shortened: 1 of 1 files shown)


Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,300 @@
...
+* Homegrown set of scripts are very clunky.
+* Maintenance of build scripts adds overhead.
+* Complete bundling of dependencies gives a much larger user installer package.
+
+
+### MacOS
+
+MacOS uses the [Homebrew](https://brew.sh/) package manager along with a custom
+set of [formulae](https://github.com/mantidproject/homebrew-mantid) for obselete
+dependencies.
```
gvardany, line 100: I would also mention that `pip` is used for all the python dependencies. This is a source of the problem, where system and python library versions can diverge. Example is mantid depending on hdf5 and h5py, where h5py also depends on hdf5, but currently we manage hdf5 via brew, h5py via pip.

gvardany, line 100: I would also mention explicitly that `pip` is used to manage all the python dependencies.

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,300 @@
...
+for RHEL 7 (and its variants CentOS, Scientfic Linux) a
+[Software Collection](https://www.softwarecollections.org/en/) is used to provide
+a gcc-7 allowing C++17-enabled builds.
+
+Pros:
+* Easier developer setup.
+* Smaller user packages. The system libraries are shared for all packages resulting
+  in a much smaller package to install.
+* (Mostly) managed for us.
+* Simply commands to install packages (apt/yum)
```
DavidFair, line 56: Nit: Simple

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,300 @@
...
+
+
+Cons:
+* Dependencies not versioned with the main code. Hard to track versions used.
+* Versions become outdated very quickly. Facilities tend to use
+  older distributions that do not have up to date versions of packages.
+* Upgrading/rebuilding package versions can cause conflicts with other packages
+  not related to mantid and even old mantid versions.
+* New dependencies have to be installed manually on developers and builders machines.
+
```
DavidFair, line 66: - User have to install Mantid repository to pull in deps, they cannot use a .deb/.rpm direct on a new machine

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,300 @@
...
+* Complete freedom to choose package versions and apply patches etc.
+* New dependencies are retrieved automatically for developers and build servers
+* Self-contained user package.
+
+Cons:
+* Dependencies not versioned with the main code. Hard to track versions used.
+* Homegrown set of scripts are very clunky.
+* Maintenance of build scripts adds overhead.
+* Complete bundling of dependencies gives a much larger user installer package.
+
```
DavidFair, line 94: - ABI changes require a complete rebuild 
- Initial download takes a significant time

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,300 @@
...
+dependencies.
+Homebrew either downloads a prebuilt binary or builds it locally from the formula.
+A developer formula is provided to easily install the required dependencies.
+
+The package provided for users contains all required dependencies
+much like Windows.
+
+Pros:
+* Build formula (mostly) maintained for us.
+* Simply commands to install packages
```
DavidFair, line 109: Nit Simple

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,300 @@
...
+
+While each approach has the drawbacks described above
+there are also general issues with the current approach:
+
+1. It is not possible to define a version of a given
+   library as the version used across the whole project.
+   This can cause confusion for developers
+   where their platform might have a newer api
+   yet their code does not work on other platforms.
+   This slows down development.
```
DavidFair, line 129: and usually ties us to the oldest LTS versions for a number of years

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,300 @@
...
+2. Using a new library takes a lot of overhead and
+   in some cases is basically impossible
+3. Three different methods requires much additional overhead
+   in understanding how things are managed.
+
```
DavidFair, line 134: 4. Pushing new dependencies to build servers requires co-ordination and potential downtime

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,300 @@
...
+Here we describe possible solutions to the problems outlined above.
+The solutions considered in detail are:
+
+* [Conan](#Conan)
+* [Conda](#Conda)
+
+Other solutions not considered:
+
+* Flatpak/Snap - would only work for Linux and still be a large effort.
+* Singularity/Docker - unsure about desktop applications & would not work for macOS.
```
DavidFair, line 167: (Docker on Windows requires the Hyper-V role last time I checked, so users would have to get admin access and have a processor capable of exposing the required instructions - so it's not universally compatible)

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,300 @@
...
+Developers would continue to work as they do now with their favourite IDEs
+and tools but would need to install Conan first.
+Release packages are [provided](https://github.com/conan-io/conan)
+that have no other dependencies and are easy to install.
+
+#### Python Packages
+
+Conan was designed as a framework to package C/C++ dependencies.
+While it is written in Python itself it does not possess the capability
+to manage Python dependencies. Windows/macOS already ship a bundled Python
```
DavidFair, line 221: Just to clarify, is it Conan ships Python on these platforms? As it reads that Windows and Mac ship Python

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,300 @@
...
+distribution so maintaining a separate site-packages is already supported here.
+It is proposed that the same be done for the desktop application bundle
+on Linux so that it is fully self-contained.
+
+The Windows/macOS Python-bundling solution is clunky.
```
DavidFair, line 226: Briefly, how? What sort of problems do we run into bundling on these platforms?

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,300 @@
...
+
+Questions: 
+
+1. Do we need to still support a build against system libraries on Linux?
+
+   * Would SNS need an RPM framework build for their GUIs linked to system Python?
+
+2. Is it sensible to consider maintaining Conan packages as suggested?
+
+  * Where would we store binary artifacts? Or just store recipes?
```
DavidFair, line 297: What would be the rough time to build locally per machine as an estimate?

If it's an hour per developer per say 6-12 months, configuring remote artefacts / storage / retention vs building locally would become a factor

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,300 @@
...
+  * Bintray has been considered but its price could be prohibitive.
+
+3. Other things not considered here?
```
DavidFair, line 300: Likely reduction or growth in binary size from these changes?

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,300 @@
...
+Python modules,
+executable programs,
+or other components"*.
+
+[Conda](https://docs.conda.io/en/latest/)
+is another cross-platform package manager but with support for a
+variety of languages including Python and C++.
+Alongside `pip` many popular Python packages support distribution through Conda.
+In fact `mantid` already has [Conda packages](https://anaconda.org/mantid/repo)
+for workbench and the framework for Linux and Mac.
```
OwenArnold, line 263: I believe the Mac packages are very stale and are effectively abandoned?

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,300 @@
...
+
+[Anaconda](https://repo.continuum.io/pkgs/) hosts many prebuilt packages for
+various combinations of compilers and operating systems.
+[Conda build](https://conda.io/projects/conda-build/en/latest/resources/commands/conda-build.html)
```
OwenArnold, line 267: Compilers are somewhat limited. Getting latest versions of compilers on all platforms is problematic/hacky at best.

OwenArnold (commented):
What is the relative priority of this work and timescales now? I appreciate this depends somewhat on the solution chosen. resourcing etc. However, I believe that this was discussed as the most important next step after 5.0 - is this still the case?

Design/ThirdpartyDependencies.md
martyngigg: It is a very good question that I have no answer for at the moment. Most core framework/qt dependencies exist in artefact repos somewhere but this is an issue as our build time is long enough.

Design/ThirdpartyDependencies.md
martyngigg: If we keep the current shared library layout then I don't see the binary size of individual shared libraries changing as we just shift what we are linking too. The package size on Linux would obvisouly grow though - this could be offset by not including ParaView any longer though...

martyngigg:
> What is the relative priority of this work and timescales now? I appreciate this depends somewhat on the solution chosen. resourcing etc. However, I believe that this was discussed as the most important next step after 5.0 - is this still the case?

Personally I think it is now the next high priority project-wide item after we drop MantidPlot as being we keep hitting issues with being on old versions of libraries. Timescales are a bit tough at the moment as they will vastly depend on the solution. If we go the suggested Conan + bundled Python route I could see this taking a good 3-6 months to test & complete properly.

martyngigg:
Thank you all for the comments so far. I've made some changes to the document in response.

My major concern still is the managment of the Python dependencies if we create a self-contained Workbench bundle - are there any specific thoughts on that part? I wonder if knocking up a prototype would be helpful?

gvardany:
I think this is the main question for everyone; this was the open point when I discussed with the conan guys in ACCU and they couldn't give a definitive answer what to do with python libraries. There is no concern regarding conan for C++ dependencies, as it promises to be cross-platform and cmake native. Personally, as a mac developer I have always been jealous of lightweight linux packages and apt package manager that handles also python dependencies. I have a feeling that this is the option minimizing the maintenance burden. I thought having some sort of brew that can handle also python dependencies just like apt, with the ability to pin minimum, maximum or exact versions would be the goal. I see an opportunity here for having a brew version/fork that is augmented to handle this, but this is out of scope for our project I guess. On the other hand, you mention that relying on system dependencies can cause conflicts with other installed software (such as octave). I think this is the case in general, and to my understanding not bundling the dependencies on linux is the standard, am I right? Could we perhaps (re)do a survey on how other similar applications handle this? Bundling python (+dependencies) in a cross-platform way is also a solution, with the downside that importing mantid from a system python might not be possible. Will this mean that the user might need to have a separate conda install of the framework side by side to the workbench app which will have its own framework instance? Also in this case, what happens if a user wants to use another python library that is not in mantid dependencies in the same workbench session with mantid, will (s)he be able to install it?

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,335 @@
...
+
+The package provided for users contains all required dependencies
+much like Windows.
+
+Pros:
+* Formula for dependencies (mostly) maintained for us.
+* Simple commands to install packages
+
+Cons:
+* Dependencies not versioned with the main code. Hard to track versions used.
```
gvardany, line 117: I would add here that for some of the dependencies we have to maintain our own formulae, such as nexusformat, and pyqt@4 in the past.

gvardany: For the record, here we could reinstate that the bundle is over 1GB and takes long to download. Waiting few minutes for download is not an issue, if it is one time thing, unfortunately it is not, since there is no mechanism of update for nightly nor release as in linux. Each time one needs a new nightly, he needs to remove the existing, download and install the new one. Of course, the bundle size will decrease when MantidPlot and Paraview are removed, but it'll still be around 500MB.

gvardany: Just thinking further on this, would it make sense to have a brew formula for mantid itself? That way one could, in principle, upgrade in-situ on command line.

Design/ThirdpartyDependencies.md
```diff
@@ -0,0 +1,335 @@
...
+
+After consideration I would suggest be that we adopt both options for supporting
+the two use cases presented. Both cases have different needs so there is no
+reason to expect one system to support both.
+
+Questions: 
+
+1. Do we need to still support a build against system libraries on Linux?
+
+   * Would SNS need an RPM framework build for their GUIs linked to system Python?
```
peterfpeterson, line 324: ORNL would prefer to connect to a `mantid-qa` package that had the extra bits of qt from mantid that we use. We would need to take inventory of what objects/subpackages we use first.

DavidFair: Could we do a `mantid-deps` and `mantid-extra-ornl` / `mantid-extra-*facility*` ...etc. similar to how the Linux kernel was packaged with Debian based systems

This way if your packaging Mantid elsewhere, it's obvious which meta packages you need. If there is enough packages, you could also have a `mantid-all` which includes all nested ones for CI jobs that can build anything?

Design/ThirdpartyDependencies.md
martyngigg: The size is a reasonable point. The question becomes though whether a smaller binary package is worth having to stick with system libraries that can become very old.

martyngigg:
> I think this is the main question for everyone; this was the open point when I discussed with the conan guys in ACCU and they couldn't give a definitive answer what to do with python libraries. There is no concern regarding conan for C++ dependencies, as it promises to be cross-platform and cmake native. Personally, as a mac developer I have always been jealous of lightweight linux packages and apt package manager that handles also python dependencies. I have a feeling that this is the option minimizing the maintenance burden. I thought having some sort of brew that can handle also python dependencies just like apt, with the ability to pin minimum, maximum or exact versions would be the goal. I see an opportunity here for having a brew version/fork that is augmented to handle this, but this is out of scope for our project I guess. On the other hand, you mention that relying on system dependencies can cause conflicts with other installed software (such as octave). I think this is the case in general, and to my understanding not bundling the dependencies on linux is the standard, am I right? Could we perhaps (re)do a survey on how other similar applications handle this?

For things shipped in standard distro repositories this is the case, yes. All libraries are separate packages installable and depending on each other. For user applications though there is a move to self-contained  packages much like we do on Windows/macOS by using things like [flatpak](https://flatpak.org/) or [snap](https://snapcraft.io/). I think the system libraries method works well for slowly changing things but for full applications it feels outdated now. Ubuntu switched their software manager to entirely use [snaps in 20.04](https://www.omgubuntu.co.uk/2020/02/ubuntu-snap-store-transition).

>Bundling python (+dependencies) in a cross-platform way is also a solution, with the downside that importing mantid from a system python might not be possible. Will this mean that the user might need to have a separate conda install of the framework side by side to the workbench app which will have its own framework instance? Also in this case, what happens if a user wants to use another python library that is not in mantid dependencies in the same workbench session with mantid, will (s)he be able to install it?

You're correct that if we also bundle python then using the system libraries is almost certainly out of the question. For a full application like workbench it feels like we should be shipping everthing we need so we are in control. A couple of examples I think of are all of the Jetbrains IDEs ship their own Java and the QtCreator standalone bundle you can download also contains the Qt libraries so that they can use the latest versions.

I agree we do have the use case of allowing users access to Python libraries that we don't ship with the bundle. The way we have been doing that so far on mac/Windows is using `pip install --user` so it goes into the user directory and the bundled Python still picks it up.

Design/ThirdpartyDependencies.md
martyngigg: @peterfpeterson Would you want the `mantid-qt` package as an rpm built against system Python? My first thought for library-based usage was conda and have 2 conda packages:
* `mantid-framework`
* `mantid-qt`

and we would not have a framework/mantidqt rpm at all.

The rpm/deb would only be used for the workbench application for ease of distribution and installation on facility systems even though that deb/rpm would contain all of the required libraries and be self-contained.

Design/ThirdpartyDependencies.md
martyngigg: A brew formula for mantid would in theory be fairly easy, especially if we didn't provide bottles so this could be an option too.

martyngigg:
Closing this. It was agreed we would move forward with [Conda](https://github.com/mantidproject/documents/blob/68a7962f07f933545d16065e0d707e4bbc97d24f/Project-Management/TechnicalSteeringCommittee/meetings/2021/2021-01-12.md)
See #97 for further discussions on packaging.

PR #85, opened 2020-11-24
Unit conversion problem in powder diffraction workflow

DannyHindson:
Initial draft of document describing changes to AlignDetectors\ConvertUnits to deal with issue raised by ISIS powder diffraction instrument scientists. I have called the document AlignDetectors_MoreRework.md so it sits next to the existing AlignDetectors_rework document from back in 2015

@peterfpeterson , @gvardany , @martyngigg - please could you have a look at this design document and let me have any comments? I've put this together to address a unit conversion problem that the ISIS powder diffraction scientists have identified. It relates to the conversions between TOF and d spacing that occur in their data reduction workflow and the way the calibration data is carried through the workflow. I would like to get this problem fixed in the 6.0 release if possible

### DIFF
--- a/Design/AlignDetectors_MoreRework.md
+++ b/Design/AlignDetectors_MoreRework.md
@@ -1,0 +1,51 @@
+Motivation
+==========
+
+The ISIS powder diffraction scientists have identified a problem with their powder diffraction reduction workflow. The workflow is as follows (algorithm names in brackets):
+
+* calibration ([CrossCorrelate](https://docs.mantidproject.org/nightly/algorithms/CrossCorrelate.html), [GetDetectorOffsets](https://docs.mantidproject.org/nightly/algorithms/GetDetectorOffsets.html)). Various approaches available here as described in the Mantid [concept page](https://docs.mantidproject.org/nightly/concepts/calibration/PowderDiffractionCalibration.html). ISIS tend to use CrossCorrelate\GetDetectorOffsets which generates a "per detector" correction in the form of a peak offset. The peak offset can easily be converted into a DIFC diffractometer constant. An alternative is the PDCalibration algorithm that directly outputs DIFC as well as DIFA\TZERO
+* convert data from TOF to d spacing ([AlignDetectors](https://docs.mantidproject.org/nightly/algorithms/AlignDetectors.html)). Uses the diffractometer constants generated in the calibration step
+* focus ([DiffractionFocussing](https://docs.mantidproject.org/nightly/algorithms/DiffractionFocussing.html)). Focus the data for each bank of detectors according to grouping data in a .cal file
+* convert data back to TOF ([ConvertUnits](https://docs.mantidproject.org/nightly/algorithms/ConvertUnits-v1.html))
+* Save data to file for downstream processing ([SaveGSS](https://docs.mantidproject.org/nightly/algorithms/SaveGSS-v1.html))
+
+I believe there is a similar workflow followed at SNS using the [AlignAndFocusPowder](https://docs.mantidproject.org/nightly/algorithms/AlignAndFocusPowder-v1.html) workflow algorithm. ISIS use some python scripts to chain together the different steps but it achieves a similar process to AlignAndFocusPowder
+
+The problem is that the final ConvertUnits step to convert from d spacing to TOF doesn't use the diffractometer constants. Following the focussing the data isn't "per detector" any more but it should still be possible to use a "per bank" DIFC\DIFA\TZERO to convert the data back to TOF
+
+There is also a more minor problem that the SaveGSS algorithm recalculates a per bank DIFC value and writes it into a header for each bank in the .gss file. This calculation also ignores the calibration output. It seems that GSAS isn't reading this header though so this problem just creates potential confusion for anyone reading the file rather than incorrect figures in GSAS.
+
+Possible solutions
+==================
+
+We could create a "backwards" mode to the AlignDetectors algorithm that performs the d spacing to TOF unit conversion including the diffractometer constants. This would then be called instead of ConvertUnits near the end of the reduction workflow.
+
+This still leaves the burden on the user\developer to remember to load the calibration data as an input parameter at two points in the workflow. Since this information has already been loaded once it seems like it should be stored in the workspace somewhere. Also the algorithm name "AlignDetectors" doesn't obviously suggest it does a unit conversion. 
+
+So on balance we prefer to put the new functionality in the ConvertUnits algorithm which is a more obvious place.
+
+Proposed Solution
+=================
+
+Modify ConvertUnits so it can incorporate diffractometer constants in its unit conversion from TOF to d spacing and vice versa.
+
+Store the "per detector" diffractometer constants in the instrument parameter map once they are applied to a workspace. A similar approach is already used to store the efixed values for indirect instruments (see ConvertUnits::getDetectorValues). This would mean that if a workspace is saved to a Nexus file, the diffractometer constants would also be saved.
+
+Create a new algorithm to load and store the diffractometer constants in the workspace called ApplyDiffCal. This would take one of three inputs:
+* calibration file (.cal or hdf)
+* a [calibration table workspace](https://docs.mantidproject.org/nightly/concepts/DiffractionCalibrationWorkspace.html)
+* an offsets table workspace
+
+When ConvertUnits does a conversion between TOF and d spacing check if the instrument map contains diffractometer constants. Use them if they're found. If none found use existing logic
+
+For unit conversions applied to spectra with multiple detectors, just use average of diffractometer constants
+
+Label the AlignDetectors algorithm as deprecated.
+
+Add some way of viewing the diffractometer data that is stored in a workspace:
+* Show the diffractometer constants in the "Show Detectors" screen?
+* Perhaps provide an algorithm to output the data to a separate table workspace eg CreateDiffCal. Although I note this would make the list of xxxDiffCal algorithms even longer
+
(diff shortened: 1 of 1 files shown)


Design/AlignDetectors_MoreRework.md
```diff
@@ -0,0 +1,51 @@
...
+* focus ([DiffractionFocussing](https://docs.mantidproject.org/nightly/algorithms/DiffractionFocussing.html)). Focus the data for each bank of detectors according to grouping data in a .cal file
+* convert data back to TOF ([ConvertUnits](https://docs.mantidproject.org/nightly/algorithms/ConvertUnits-v1.html))
+* Save data to file for downstream processing ([SaveGSS](https://docs.mantidproject.org/nightly/algorithms/SaveGSS-v1.html))
+
+I believe there is a similar workflow followed at SNS using the [AlignAndFocusPowder](https://docs.mantidproject.org/nightly/algorithms/AlignAndFocusPowder-v1.html) workflow algorithm. ISIS use some python scripts to chain together the different steps but it achieves a similar process to AlignAndFocusPowder
+
+The problem is that the final ConvertUnits step to convert from d spacing to TOF doesn't use the diffractometer constants. Following the focussing the data isn't "per detector" any more but it should still be possible to use a "per bank" DIFC\DIFA\TZERO to convert the data back to TOF
```
peterfpeterson, line 14: `AlignAndFocusPowder` uses an effective detector position to convert back to TOF (if supplied) or the center-of-mass of the pixels contributing to the focused spectrum. The effective postions are specified using `PrimaryFlightPath`, `L2`, `Polar`, and `Azimuthal` parameters of `AlignAndFocusPowder`/`AlignAndFocusPowderFromFiles`. Not specifying these values will give the default behavior (center of mass position) which is essentially nonsense for a 2nd generation (and beyound) powder instrument.

Design/AlignDetectors_MoreRework.md
```diff
@@ -0,0 +1,51 @@
...
+
+There is also a more minor problem that the SaveGSS algorithm recalculates a per bank DIFC value and writes it into a header for each bank in the .gss file. This calculation also ignores the calibration output. It seems that GSAS isn't reading this header though so this problem just creates potential confusion for anyone reading the file rather than incorrect figures in GSAS.
```
peterfpeterson, line 16: Essentially the same as the previous comment. The DIFC was added to the header to help people audit which instrument parameter file, `.iparm`, is the correct one to use with the data. It is also used for people that want to "write a python script to load and plot in d-spacing" without mantid. Since this algorithm can be used as a component of total scattering, this is a useful feature.

Design/AlignDetectors_MoreRework.md
```diff
@@ -0,0 +1,51 @@
...
+So on balance we prefer to put the new functionality in the ConvertUnits algorithm which is a more obvious place.
+
+Proposed Solution
+=================
+
+Modify ConvertUnits so it can incorporate diffractometer constants in its unit conversion from TOF to d spacing and vice versa.
+
+Store the "per detector" diffractometer constants in the instrument parameter map once they are applied to a workspace. A similar approach is already used to store the efixed values for indirect instruments (see ConvertUnits::getDetectorValues). This would mean that if a workspace is saved to a Nexus file, the diffractometer constants would also be saved.
+
+Create a new algorithm to load and store the diffractometer constants in the workspace called ApplyDiffCal. This would take one of three inputs:
```
peterfpeterson, line 34: All 3 of these should, at least internally, get converted to [diffraction calibration workspace](https://docs.mantidproject.org/nightly/concepts/DiffractionCalibrationWorkspace.html).

peterfpeterson:
Lots of comments on this one, maybe I've been here too long. I think most of this can be attributed to not fully understanding all of the parameters in `AlignAndFocusPowder`.
1. Using the properties `PrimaryFlightPath`, `L2`, `Polar`, and `Azimuthal` sets the focus position of all of the output spectra via [EditInstrumentGeometry](https://docs.mantidproject.org/nightly/algorithms/EditInstrumentGeometry-v1.html). The resulting workspace will work exactly as expected in [ConvertUnits](https://docs.mantidproject.org/nightly/algorithms/ConvertUnits-v1.html). The `SpectrumIDs` parameter is the other confusing one that may be easier to understand by reading the code.
1. "Offsets" are very useful for seeing how much parameters are changed, but are relative to a previous instrument geometry. We had written some code once that allowed for calculating offsets between any two calibrations `(new - old)/old` which was fairly useful for seeing how the instrument evolved over time. The change of `AlignAndFocusPowder` to use DIFC was to break this dependency and allow for new methods of calibration which can start from a previous calibration. [PDCalibration](https://docs.mantidproject.org/nightly/algorithms/PDCalibration-v1.html) was designed with this new thinking.
2. `DIFC ~ L2 sin(theta)` cannot be converted back to a detector position. [AlignComponents](https://docs.mantidproject.org/nightly/algorithms/AlignComponents-v1.html) will move instrument components around until the DIFC from the detector position is as close as possible to those from calibration.
2. Having DIFC available in the instrument view is an interesting extra bit of information.
1. Using `ConvertUnits` at the end of `AlignAndFocusPowder` will use the same "wrong geometry" to go back to d-spacing. Then one can use `EditInstrumentInfo` and convert back to TOF. Again, the extra parameters on `AlignAndFocusPowder` do this already.

Design/AlignDetectors_MoreRework.md
DannyHindson: > Lots of comments on this one, maybe I've been here too long. I think most of this can be attributed to not fully understanding all of the parameters in `AlignAndFocusPowder`.
> 
> 1. Using the properties `PrimaryFlightPath`, `L2`, `Polar`, and `Azimuthal` sets the focus position of all of the output spectra via [EditInstrumentGeometry](https://docs.mantidproject.org/nightly/algorithms/EditInstrumentGeometry-v1.html). The resulting workspace will work exactly as expected in [ConvertUnits](https://docs.mantidproject.org/nightly/algorithms/ConvertUnits-v1.html). The `SpectrumIDs` parameter is the other confusing one that may be easier to understand by reading the code.
> 2. "Offsets" are very useful for seeing how much parameters are changed, but are relative to a previous instrument geometry. We had written some code once that allowed for calculating offsets between any two calibrations `(new - old)/old` which was fairly useful for seeing how the instrument evolved over time. The change of `AlignAndFocusPowder` to use DIFC was to break this dependency and allow for new methods of calibration which can start from a previous calibration. [PDCalibration](https://docs.mantidproject.org/nightly/algorithms/PDCalibration-v1.html) was designed with this new thinking.
> 3. `DIFC ~ L2 sin(theta)` cannot be converted back to a detector position. [AlignComponents](https://docs.mantidproject.org/nightly/algorithms/AlignComponents-v1.html) will move instrument components around until the DIFC from the detector position is as close as possible to those from calibration.
> 4. Having DIFC available in the instrument view is an interesting extra bit of information.
> 5. Using `ConvertUnits` at the end of `AlignAndFocusPowder` will use the same "wrong geometry" to go back to d-spacing. Then one can use `EditInstrumentInfo` and convert back to TOF. Again, the extra parameters on `AlignAndFocusPowder` do this already.

Thanks a lot for taking a look at this.

I've had another look at the AlignAndFocusPowder algorithm and I see the EditInstrumentGeometry step that I'd missed before and I can see that would solve the problem if you have a suitable effective position for each focussed group. The ISIS powder calibration workflow doesn't involve producing an effective position however. We're just using the offsets approach to represent the calibrated detector positions vs the IDF as a peak offset. So the problem I'm trying to fix may be ISIS-specific. I guess we could change the workflow to be based around effective detector positions but in my discussions with the scientists here I had the impression this could be fixed without a change in the calibration approach.

ISIS don't use AlignAndFocusPowder but we do run a similar chain of algorithms. I think the scenario that the ISIS instruments use is closest to running AlignAndFocusPowder with the CalibrationWorkspace parameter supplied (and no effective positions supplied). In this case the final conversion from d spacing to TOF looks to me as though it ignores the DIFCs in the CalibrationWorkspace and just uses a DIFC based on the nominal detector positions from the IDF. It seems like ConvertUnits should use the DIFCs from the CalibrationWorkspace during the final conversion to TOF in this case?

I don't think the changes I'm proposing would get in the way of a workflow based on effective positions. I would blank out any DIFCs in the instrument parameter map whenever any of the detector positions are changed eg if EditInstrumentGeometry is run. So ConvertUnits would do what it's always done in your workflow even if I make these changes.

DannyHindson:
@gvardany - do you know if this topic is relevant to the powder diffraction instruments at ILL? It's problem relating to the unit conversions that happens in the ISIS powder diffraction data reduction process. The unit conversions are from TOF to d spacing (pre-focussing) and back to TOF (after focussing) and the problem relates to whether the "per detector" calibration data is used in the final unit conversions.
I had a look at the PowderILLDetectorScan and PowderILLParameterScan algorithms and it doesn't look like the ILL instruments follow the same reduction process so perhaps not relevant to you but thought I'd check. The nature of the D2B\D20 instruments may mean you don't need to follow the same calibration process.

gvardany:
ILL is not concerned about TOF diffraction.

DannyHindson:
> ILL is not concerned about TOF diffraction.

OK yes of course. Thanks for clarifying

DannyHindson:
@peterfpeterson - I think we got to the point in our offline conversation at the start of the week where you're happy for me to proceed with this change (subject to normal PR approval etc). If yes, please can you add a comment on here and then I'll ask Martyn to approve this PR to add the doc to the designs folder. I've just updated the doc slightly with a couple of comments from our discussion.

peterfpeterson (approved):
The design looks reasonable. Hopefully there isn't a performance hit on unit conversion.

PR #88, opened 2021-02-22
Design document for pre-commit

Pasarus:


Design/Pre-commit/Pre-commit implementation detail.md
```diff
@@ -0,0 +1,110 @@
...
+All final tasks should be completed in one day on to minimise potential disruption to developers in the team 
+- Update and apply clang-format and Yapf versions to the whole codebase, using the pre-commit hooks 
+- Replace Jenkins static analysis jobs with the already made pre-commit CI job 
+- Inform developers
+
+## Potential future feature plans
+Future development will be easy and not require significant updates to CMake for Windows or require custom installation, updating the .yaml config will allow future changes to not only the local developer machines but on Jenkins seamlessly.
+
+- Any newly added feature will be run on the whole repo at the same time to avoid random things being added to new PRs and ideally should be enforced by Jenkins, but many suggested features won't require jenkins enforcement. 
+- Support more things in pre-commit slowly introducing helpful features
```
DanNixon, line 94: Add key and password finding to this list also, on the off chance someone thinks that is a good idea.

peterfpeterson:
The purpose of pre-commit is to prevent developers from pushing "bad code" that simple checks before committing would have caught. Whatever is implemented should assume that build servers have the final say in whether the code is "acceptable" according to automated tools. I prefer not having pre-commit implement our entire build-system locally. I agree that formatting checks (e.g. clang-format, yapf) and some amount of static analysis are useful, but I'd rather leave fuller checks out of it.

An important part of the conversation is what type of checks should be done. The mechanics of adding things to pre-commit is easy enough where we could easily go overboard quickly.

Design/Pre-commit/Pre-commit implementation detail.md
Pasarus: Thought I'd added this, I am keenly aware of the issues with key and password leakage having done that myself on a personal project (Never again!)

martyngigg (approved):
TSC were happy with the what was discussed.

Approved in meeting on [2021-03-02](https://github.com/mantidproject/documents/blob/7168e89ca1352d29476debd81b3f03d3a453762d/Project-Management/TechnicalSteeringCommittee/meetings/2021/2021-03-02.md)

PR #89, opened 2021-03-08
Add design doc for muon fitting widgets

robertapplin:
Add design doc for muon fitting widgets

To test read through the design doc and make sure it makes sense.


martyngigg:
@robertapplin Thanks for the design document here. It's hard to review a word doc properly in git and it also makes it hard to see it back in the history. Any chance you could extract it to markdown? If you have pictures etc then there is an example in mantidproject/documents#49 of how this can be done.

robertapplin:
Thanks @martyngigg , I have converted it to markdown now

PR #96, opened 2021-06-11
Initial SXD workflow and requirements

RichardWaiteSTFC:
Powerpoint of SXD workflow and requirements after discussion with Silvia and Alberto


DannyHindson:
@martyngigg - Richard has created this Powerpoint describing the outstanding SXD work that we have to do. We've sent it to Silvia for feedback but thought I'd share with you as well

PR #97, opened 2021-07-16
Conda package implementation

Pasarus:
This is a proposal for future packaging practices given the planned move to a Conda-first strategy


Design/Packaging/Future_packaging_strategy.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Intended move forward
+
+All operating systems should have a fully built and deployable conda-build implementation, where a use can in theory run a simple command in conda and it will install Mantid with all it's runtime dependencies, which with some small tweaks we can then package on all Operating Systems assuming that the conda-build recipe has been created and ran successfully.
+
+The aim at this time is to create similar packages to what we have now on MacOSX and Windows, whilst shipping all dependencies in our Ubuntu and CentOS/RHEL7 which is a change. We will make a .exe for Windows, .dmg for MacOSX, .deb for Ubuntu, and a .rpm for CentOS/RHEL7 much like we do now on all operating systems.
+
+### Conda build
+
+We intend to distribute a conda package using the conda-build system, which will allow advanced users to install Mantid using conda. Users will be able to install this packaged version with,
```
martyngigg, line 22: Can we include here that we intend to build separate conda packages for::
 - `mantid-framework` (as we already do)
 - `mantid-qt`
 - `mantid-workbench`

The first two are libraries that other applications can reuse.

Design/Packaging/Future_packaging_strategy.md
```diff
@@ -0,0 +1,70 @@
...
+```bash
+conda create -p package_dir mantid-workbench --copy -c mantid -c conda-forge -y
+```
+
+In the case of a package that was not uploaded to the Anaconda Cloud like the previous command expects, you can do this from a locally constructed package and achieve the same result using this command:
+```bash
+conda create -p package_dir --use-local mantid-workbench --copy -c conda-forge -y
+```
+
+For Linux operating systems I suggest using FPM (`Effin Package Management - https://github.com/jordansissel/fpm) due to it's simplicity and ease of creating .deb and .rpm installers, from one machine, therefore in the future it should be possible to create all Linux distribution packages from one machine. At present it is not planned to use FPM for MacOSX as it doesn't support .dmg app bundles, and the support it does provide for .pkg files on MacOSX is less than ideal for the same reasons we already migrated away from .pkg files in the first place.
```
martyngigg, line 46: I there might be a backtick missing here?

Design/Packaging/Future_packaging_strategy.md
```diff
@@ -0,0 +1,70 @@
...
+
+We intend to distribute a conda package using the conda-build system, which will allow advanced users to install Mantid using conda. Users will be able to install this packaged version with, 
+```bash
+conda install -c mantid mantid-workbench
+```
+The user will then be able to interact with the package in a manner familiar to any users of other conda packages, e.g. Spyder, Numpy,
+```bash
+conda install -c mantid mantid-workbench // Install conda package
+python -c "import mantid"  // Import mantid
+MantidWorkbench // Launch workbench
```
peterfpeterson, line 30: I'm certain that some people will disagree with pascal case for launching the program. Especially on linux.

peterfpeterson (commented):
A couple things that should be decided:
1. Somewhere in all of the conda work we should have a description of versioning/labels. Most of the `mantid-framework` packages are "nightly" but very few have actual labels to guide the user to the version they are actually looking for.
2. Will we support multiple versions of python (e.g. 3.6 and 3.8)?
2. Will we support multiple versions of numpy within a single version of python? Other packages that are fundamental enough should be called out as well. My inclination is that we should only have one version per version of python.
3. How do we work through thinlinc?

Design/Packaging/Future_packaging_strategy.md
SimonHeybrock: Another thing to potentially consider here (something we ran into with scipp, and haven't fully decided on a solution yet):

How should optional dependencies be handled? Not sure what a good example of this would be in Mantid, but I am thinking of things that may be required to use certain features (in scipp plotting is an example: You don't really need matplotlib if you just want to use the lib, but if you want to plot, you do). Would you create meta packages such as `mantid-minimal` and `mantid-all`?

OwenArnold:
> 1. Somewhere in all of the conda work we should have a description of versioning/labels. Most of the `mantid-framework` packages are "nightly" but very few have actual labels to guide the user to the version they are actually looking for.

Shouldn't the use of conda channels be enough?

Design/Packaging/Future_packaging_strategy.md
Pasarus, line 46:
```suggestion
For Linux operating systems I suggest using FPM (Effin Package Management - https://github.com/jordansissel/fpm) due to it's simplicity and ease of creating .deb and .rpm installers, from one machine, therefore in the future it should be possible to create all Linux distribution packages from one machine. At present it is not planned to use FPM for MacOSX as it doesn't support .dmg app bundles, and the support it does provide for .pkg files on MacOSX is less than ideal for the same reasons we already migrated away from .pkg files in the first place.
```

Pasarus: That tick itself shouldn't have been there, it has been removed thanks for spotting this!

Design/Packaging/Future_packaging_strategy.md
Pasarus: > Can we include here that we intend to build separate conda packages for::
> 
>     * `mantid-framework` (as we already do)
> 
>     * `mantid-qt`
> 
>     * `mantid-workbench`
> 
> 
> The first two are libraries that other applications can reuse.

I didn't know we were looking at producing a `mantid-qt` package, perhaps we should discuss scope/whether or not this is expected as part of the current effort towards conda-ification.

Pasarus: > Another thing to potentially consider here (something we ran into with scipp, and haven't fully decided on a solution yet):
> 
> How should optional dependencies be handled? Not sure what a good example of this would be in Mantid, but I am thinking of things that may be required to use certain features (in scipp plotting is an example: You don't really need matplotlib if you just want to use the lib, but if you want to plot, you do). Would you create meta packages such as `mantid-minimal` and `mantid-all`?

I think I discussed this briefly in the Future looking part of the document. For now, my understanding is that we do not want to support optional dependencies, as for most of the dependencies they are absolutely required. Right now we are producing a `mantid-all` effectively. I think long term it is very possible to make `mantid-minimal` through to some midpoint for example `mantid-muon` to `mantid-all` where only the absolutely needed dependencies are pulled in.

Design/Packaging/Future_packaging_strategy.md
Pasarus: What would you suggest as an alternative? This was just an example I came up with at the time. Would `mantid-workbench` or `mantidworkbench` work better do you think?

Design/Packaging/Future_packaging_strategy.md
```diff
@@ -0,0 +1,70 @@
...
+## Alternatives
+
+The various Linux distributions have their built in .deb and .rpm creation tools, as well as the abstraction layer that comes from CPack enabling building and linking to system packages. We do not want to link to the system packages, whilst conda-build will get all the required packages for us whilst CPack is not capable of this.
+
+MacOSX currently uses CPack, it has a custom ruby script and would likely work for our current purposes, however it is frustrating to maintain, we would be able to unite the script for package creation for MacOSX and POSIX systems and a lot of the details should allow for easier packaging scripts. 
+
+Windows at present uses CPack and NSIS to create the packaging. With CPack being rolled back, we intend to stick with NSIS via the larger packaging script. At present we just put in the required dependency .dll by installing them into the bin of the package, Conda will handle this for us.
+
+## Looking to the future
+Considering Mantid workbench is now approaching behemouth status like Mantid Plot, it may be prudent to look at potential future project deliverables that don't include all parts of workbench that are included now.
```
DavidFair, line 66: More of a nit, but I'm not sure on calling Workbench a behemoth like Plot, as Plot was unmaintainable from no tests and god classes. Workbench, which is large but well tested, isn't a "behemoth" and might give the wrong ideas to readers. 

Could we use something like, "Mantid Workbench is large and all-encompassing" or similar?

Design/Packaging/Future_packaging_strategy.md
```diff
@@ -0,0 +1,70 @@
...
+```bash
+conda create -p package_dir mantid-workbench --copy -c mantid -c conda-forge -y
+```
+
+In the case of a package that was not uploaded to the Anaconda Cloud like the previous command expects, you can do this from a locally constructed package and achieve the same result using this command:
+```bash
+conda create -p package_dir --use-local mantid-workbench --copy -c conda-forge -y
+```
+
+For Linux operating systems I suggest using FPM (Effin Package Management - https://github.com/jordansissel/fpm) due to it's simplicity and ease of creating .deb and .rpm installers, from one machine, therefore in the future it should be possible to create all Linux distribution packages from one machine. At present it is not planned to use FPM for MacOSX as it doesn't support .dmg app bundles, and the support it does provide for .pkg files on MacOSX is less than ideal for the same reasons we already migrated away from .pkg files in the first place.
```
DavidFair, line 46: Could we also add:

- A link to something why we don't use .pkg files (just a simple link will do)
- A note stating we can write one script for RPM and DEB packages at the same time

Design/Packaging/Future_packaging_strategy.md
martyngigg: > > Can we include here that we intend to build separate conda packages for::
> > ```
> > * `mantid-framework` (as we already do)
> > 
> > * `mantid-qt`
> > 
> > * `mantid-workbench`
> > ```
> > The first two are libraries that other applications can reuse.
> 
> I didn't know we were looking at producing a `mantid-qt` package, perhaps we should discuss scope/whether or not this is expected as part of the current effort towards conda-ification.

Things like [mslice](https://github.com/mantidproject/mslice), [PyRs](https://github.com/neutrons/PyRs), [addie](https://github.com/neutrons/addie) use features from the `mantidqt` library but are separate and to do not need workbench so it is beneficial to have them separated.

martyngigg: > Another thing to potentially consider here (something we ran into with scipp, and haven't fully decided on a solution yet):
> 
> How should optional dependencies be handled? Not sure what a good example of this would be in Mantid, but I am thinking of things that may be required to use certain features (in scipp plotting is an example: You don't really need matplotlib if you just want to use the lib, but if you want to plot, you do). Would you create meta packages such as `mantid-minimal` and `mantid-all`?

We don't have a good solution to this either to be honest. We have a few algorithms that can use dependencies if they are available and it would be nice to to have to include everything as a hard dependency. We probably have a worse problem than `scipp` in terms of the number of ways we can carve up the codebase so what future meta-packages would be is challenging. Were you thinking of going to the level of a `scipp-core` and then separate facility packages?

Design/Packaging/Future_packaging_strategy.md
martyngigg: I think we can work this out later, right? We use PascalCase on Windows/MacOS but we can keep all lower case on linux.

Design/Packaging/Future_packaging_strategy.md
SimonHeybrock: Since you last saw `scipp` we have actually removed anything neutron-specific form it, so what I am mentioning above is even more generic dependencies. There is now a separate `scippneutron`. The latter does (and will not) not contain facility-specifics either, those have to be in separate packages (we have `ess` for this).

Design/Packaging/Future_packaging_strategy.md
martyngigg: I agree. The core functionality of Workbench is far reduced from MantidPlot and Workbench is far more modular and maintainable. I see the all-encompassing thing as the `-all` package being a bit like `Anaconda` itself - it contains everything that you might need and it easy for users to know what they are getting. There are also good reasons for providing the other, more minimal, package types but it's not because Workbench  itself is too big.

Design/Packaging/Future_packaging_strategy.md
peterfpeterson: We have packages that would use `mantid-qt`, but could require `mantid-workbench` instead. That is essentially what we did long ago.

peterfpeterson:
> > 1. Somewhere in all of the conda work we should have a description of versioning/labels. Most of the `mantid-framework` packages are "nightly" but very few have actual labels to guide the user to the version they are actually looking for.
> 
> Shouldn't the use of conda channels be enough?

[Labels](https://anaconda.org/mantid/mantid-framework/labels) are probably a better way to do it. For framework, the labels essentially separate nightly and stable releases. It allows users to install / update to the latest stable version by adding a label. 

A [channel](https://anaconda.org/mantid) is more in line with a project/entity.

However, [pytorch](https://anaconda.org/search?q=pytorch) seems to do it both ways.

Design/Packaging/Future_packaging_strategy.md
```diff
@@ -0,0 +1,70 @@
...
+
+For MacOSX, we need to make .dmg files, we can use a very similar strategy to linux here but with hdiutil (Disk Utility built into MacOS used similarly to how FreeCAD has: https://github.com/FreeCAD/FreeCAD-Bundle/blob/master/conda/osx/create_bundle.sh)
+
+For Windows, we intend to continue using NSIS to produce a .exe installer. We evaluated WIX unfortunately the amount of time it would take to not only switch but to learn how to switch from NSIS is not worth the benefits, this isn't out of the question long term. 
+
+
+## An overall look at the process
+Everything will be achieved from a single script that can be worked into CMake or some other method of calling. The packaging script should return either a locally built Conda-build that is then uploaded or not uploaded, depending on this upload status we then create the environment by either downloading the dependencies and mantid workbench, or just the dependencies using a local mantid workbench constructed with Conda-build. I have produced a Sequence Diagram to better explain the suggested process:
```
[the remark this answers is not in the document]
peterfpeterson: Analogous to packaging for conda (which I understand much better) is `.spec` for generating rpms. CMake can make rpms without much intervention, but it is much better to hand write a `.spec` file that builds using CMake, and configures the install/packaging steps itself. The `.spec` "language" has facilities to make this fairly straightforward and I believe conda does as well. There will be some judgement applied on where to put some parts of the configuration in this process.

Design/Packaging/Future_packaging_strategy.md
Pasarus: In response, @StephenSmith25 has put together a plan to move towards `MantidQt` being independent of workbench, and dependent on `MantidFramework` in order to compile from `MantidFramework` some work will need to be done to ensure that this package provides the correct headers. Then the end result of all of this a `MantidWorkbench` package will be produced that is dependent on `MantidFramework` and `MantidQt` and will depend on the available headers that these provide at compile time for Conda Build. This is my understanding of how work is progressing at present, I will put something in the document that speaks to this effect.

Pasarus:
> A couple things that should be decided:
> 
>     1. Somewhere in all of the conda work we should have a description of versioning/labels. Most of the `mantid-framework` packages are "nightly" but very few have actual labels to guide the user to the version they are actually looking for.
> 
>     2. Will we support multiple versions of python (e.g. 3.6 and 3.8)?
> 
>     3. Will we support multiple versions of numpy within a single version of python? Other packages that are fundamental enough should be called out as well. My inclination is that we should only have one version per version of python.
> 
>     4. How do we work through thinlinc?

1. I will add something to the document discussing labels, the general consensus was to follow what similar packages have done in regards to labeling where stable versions will be labelled with `main`, unstable versions with `dev` and for our nightly builds, add a `nightly` label.

2. In terms of python support, we should come up with a minimum support schema, python itself supports python versions with bug fixes for up to 18 months after final release. I would suggest something a long the lines of 
```
18 months after a release of a python version that will become our minimum version and any prior versions will be dropped from support
```
therefore in this case our minimum version would be 3.8 and 3.6/3.7 would be dropped from Mantid, with an aim to support 3.9 when possible. This keeps us roughly in-line with the python language, we would never necessarily lose support for older versions when support is dropped, it's just support for newer versions of Mantid; as the Anaconda cloud will be able to supply older versions for some time after.

3. With numpy/boost/gsl etc We need to be careful as to which versions we do and don't support, we can have an idea of a range in where mantid might work, but we shouldn't really be shipping those. One suggestion is similar with how python support would be dictated by an 18 month window, we only support versions that came out in the last 18 months, for example right now that would be numpy 1.19, with an aim to bump numpy versions every time we bump python versions. Any specifics will need to be discussed and explicitly agreed upon and I am not sure that's what this document is for, as it mostly aims at solving the way we package and how we resolve dependencies not what these dependencies explicitly are. There is always the option of piggybacking on Conda-forge until we hit a problem that stops us from doing so, where by we use their min versions which are numpy 1.19.5 and python 3.8 for example.

4. With regards to thinlinc I don't quite understand how that operates and I would like to talk through with you any misgivings, in terms of CentOS7/RHEL7 we would be using a .rpm that ships all dependency libraries inside of it, so I don't think it will be an issue? If you need to go over any caveats I am happy to arrange a meeting with you to discuss this @peterfpeterson

Design/Packaging/Future_packaging_strategy.md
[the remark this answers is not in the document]
Pasarus: The likelihood is, given Windows is a pain when it comes to playing nice with cross platform things, that we will have a POSIX script and a Windows scripts, this is implementation detail however. I am against building a big god like script similar to our present jenkins buildscript. However, I do not think this script will get very large a few hundred lines at most, that's with multiple issues that I currently am not aware of.

These scripts could be called via CMake and perform the packaging role for us, it will be a bit of a step back in terms of our current in built CPack/CMake support but there isn't a better solution that I am aware of. The best we can do here, as far as I know, is mimic how we do it now utilizing a bash and batch script.

Design/Packaging/Future_packaging_strategy.md
Pasarus: Yes that makes more sense to me I will change it.

Design/Packaging/Future_packaging_strategy.md
Pasarus: I will add that to the todo list.

Design/Packaging/Future_packaging_strategy.md
```diff
@@ -0,0 +1,70 @@
...
+
+- On CentOS/RHEL7, we create a package containing only our compiled program in a .rpm format. We rely on the currently available system packages, optional packages from `rhel-7-workstation-optional-rpms` and some packages from EPEL to run Mantid. We install the Mantid .rpm package using the rpm command.
+
+## Intended move forward
+
+All operating systems should have a fully built and deployable conda-build implementation, where a use can in theory run a simple command in conda and it will install Mantid with all it's runtime dependencies, which with some small tweaks we can then package on all Operating Systems assuming that the conda-build recipe has been created and ran successfully.
+
+The aim at this time is to create similar packages to what we have now on MacOSX and Windows, whilst shipping all dependencies in our Ubuntu and CentOS/RHEL7 which is a change. We will make a .exe for Windows, .dmg for MacOSX, .deb for Ubuntu, and a .rpm for CentOS/RHEL7 much like we do now on all operating systems.
```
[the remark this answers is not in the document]
Pasarus: Added to the list of things to do

peterfpeterson:
The thinlinc stuff is with [vglwrapper here](https://github.com/mantidproject/mantid/blob/master/buildconfig/CMake/LinuxPackageScripts.cmake#L148-L160). We need to replicate that functionality somehow for opengl to work remotely.

Pasarus:
I think I have addressed everything except thinlinc now in the document or directly.

With regards to thinlinc as it seems to be implementation detail, I will leave it out, but I don't see why an executable of ours on Linux can't support it like it does now.

Design/Packaging/Future_packaging_strategy.md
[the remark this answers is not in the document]
martyngigg: I think we can work out having the current `package` target do what we need so that `cmake --build . --target package` would work out the right thing to do for us? In that respect I'm not sure it's a step back. The CPack generation of debs/rpms has always been just "good-enough" anyway and there is alot of the install packaging handling that will be cleaned up by conda anyway. 

As Sam linked to, [FreeCad](https://github.com/FreeCAD/FreeCAD-Bundle/tree/master/conda) simply has a set of directories for each of the platforms containing the packaging scripts. If needed I think we can extract out common features into "script modules" used by each?

Are people happy that we use `fpm` to produce the rpms/debs rather than writing spec/control files by hand?

martyngigg:
Along with vglwrapper we also need to make sure we capture `jemalloc` in the runtime requirements for workbench.

PR #113, opened 2023-09-06
Fix error in plotting smoke testing instructions

jhaigh0:
In response to https://github.com/mantidproject/mantid/issues/35752

 - Changed `LOQ7014` to `LOQ74014`

The python script makes refference to this spread sheet having an OS column in order to make different issues for each OS listed. This column doesn't appear to be in the template? Are they just added when making the issues or am I looking in the wrong place?


sf1919:
The first sheet of this excel spredsheet contains the OS information. The script cycles through each of the OS for each of the issues and ensures that a separate issue is raised for each OS with the information specific to that OS.

sf1919 (approved):
The changes are what were asked for. Many thanks!!!

PR #141, opened 2026-09-21
Add powder normalization design

AndreiSavici:
Add design for implementing MDNorm for powders


mducle (approved):
Looks good to me. Just a couple of nitpicking points:

* Missing close bracket in citation before eq (1)
* Section 2.1, p3, para 4 - "if the discriminant is negative, there are no solutions" - I think this also means that Q does not satisfy the conservation laws and is not physically measurable.
* Section 3, p5, point 1 - "integrated into a single bin" -> "integrated into a single energy bin" (for clarity, I know there's only one dimension of energy but...)
* Section 3, p5, point 5 - change "intensity" -> "normalization factor" (two occurrences)

